PR #23, opened 2022-03-17
Raise on unknown perf event kind

cgaebel:
This PR makes magic-trace raise when it sees an unrecognized perf event kind instead of silently ignoring it.

Unrecognized perf events are bugs. Silently ignoring them was a fine idea during initial devleopment of magic-trace, but has quickly fallen off as an optimal strategy as we try to fix the long tail of issues in
different environments.


Xyene (approved):
LGTM, thanks!

PR #27, opened 2022-03-17
Support PIDs and TIDs of 0

cgaebel:
Perf uses PID=0 or TID=0 to attach no PID or TID to an event. When it does that, make magic-trace treat them as optional. When it actually goes to write a trace, it now writes a PID/TID of '?'.

Previous behavior was to crash on Pid.of_int if the int was 0.


Xyene (approved):
LGTM, thanks!

PR #28, opened 2022-03-17
Remove the `Unknown` variant for events

cgaebel:
All events are known. All events must be known. If we don't know an event, it's a bug.


Xyene (approved):
LGTM, thanks!

PR #42, opened 2022-03-18
Record event's ip address, symbol, and offset

gretay-js:
This info makes it easier to use magic trace events for feedback-directed optimizations.


src/perf_tool_backend.ml
```diff
@@ -104,13 +104,23 @@ let finish_recording { pid; _ } =
...
       " %d/%d %d.%d: %s@=> %Lx %s@$"
-      (fun pid tid time_hi time_lo kind addr rest ->
+      (fun pid tid time_hi time_lo kind_and_ip addr rest ->
+        let kind_and_ip, ip_rest = String.rsplit2_exn ~on:' ' (String.strip kind_and_ip) in
+        let kind, ip_string = String.rsplit2_exn ~on:' ' (String.strip kind_and_ip) in
+        let ip =
+          try Scanf.sscanf ip_string "%Lx" (fun ip -> ip) with
+          | Scanf.Scan_failure _ | End_of_file -> 0L
+        in
+        let ip_symbol, ip_offset =
```
Xyene, line 120: Nit: this bit of parsing is identical to the `symbol, offset` one below, could we add a small local parsing helper here?

It's a little sad that the `perf` output doesn't admit easier parsing here, but what can you do...

Xyene (commented):
Otherwise LGTM (also lint and DCO checks)

src/perf_tool_backend.ml
gretay-js: Fixed and the CI is green.

PR #47, opened 2022-03-19
Automatically load perf map files

### LINKED ISSUE #4: Any support planned for managed languages?
From what I can tell magic-trace right now only works with native binaries as you need to select a symbol address up front.

Do guys have any plans for managed languages? Maybe via the perf symbol map files?

cgaebel:
This adds support for automatically loading perf map files,
documented [here](https://github.com/torvalds/linux/blob/master/tools/perf/Documentation/jit-interface.txt).

I tested it on a C# and java program. The traces are still a little
messed up, but the perf map file is clearly being used to resolve
symbols. For full JIT support, we will likely need help from people
actually familiar with how each JIT works.

Fixes #4. Not because we actually added support for managed languages,
but because this is the end of general support for managed languages
and we'll need separate bugs to track each managed language people want
to use.


cgaebel:
Before:
Screen Shot 2022-03-19 at 9 36 55 AM

After:
Screen Shot 2022-03-19 at 1 37 33 PM

src/perf_tool_backend.ml
```diff
@@ -106,43 +106,65 @@ module Perf_line = struct
...
+    | "tr end" -> [ End None ]
+    | "tr end  call" -> [ End Call ]
+    | "tr end  return" -> [ End Return ]
+    | "tr end  syscall" -> [ End Syscall ]
+    | "jmp" -> [ Jump ]
+    | "jcc" -> [ Jump ]
+    (* CR cgaebel: I have no idea if these are right but they led to the least broken traces out
+       of the options I tried. I didn't try many. *)
+    | "syscall" -> [ Call ]
+    | "int" -> [ Call ]
```
Xyene, line 129: Played around with this a bit. I'm not sure what V8 is using interrupts for, but modelling them as jumps rather than calls makes the traces look a lot better:

image

cgaebel: I'm going to add software interrupts in a separate feature.

core/perf_map_location.ml
```diff
@@ -0,0 +1,11 @@
+open! Core
+
+type t =
+  { start_addr : int64
```
Xyene, line 4:
```suggestion
  { start_addr : Int64.Hex.t
```

for the `sexp_of_t`

Xyene (commented):
Typo in commit title: "Support perf map flies"

src/perf_tool_backend.ml
```diff
@@ -181,16 +187,25 @@ module Perf_line = struct
         in
         let src_instruction_pointer = int64_of_hex_string src_instruction_pointer in
         let dst_instruction_pointer = int64_of_hex_string dst_instruction_pointer in
-        let parse_symbol_and_offset str =
+        let parse_symbol_and_offset str ~addr =
           match Re.Group.all (Re.exec symbol_and_offset_re str) with
-          | [| _; symbol; offset |] -> symbol, Int.Hex.of_string offset
-          | _ | (exception _) -> "[unknown]", 0
+          | [| _; symbol; offset |] -> Symbol.From_perf symbol, Int.Hex.of_string offset
```
Xyene, line 192: It's a little strange that `perf` isn't resolving these symbols :)

Let's leave a comment here to remind ourselves of this weirdness later.

cgaebel: Done

core/perf_map.ml
```diff
@@ -0,0 +1,170 @@
...
+3a6caa5142e 18b LazyCompile:~inspect node:internal/util/inspect:292
+3a6caa51926 1cf LazyCompile:~formatValue node:internal/util/inspect:745
+3a6caa5241e cfd LazyCompile:~formatRaw node:internal/util/inspect:820
+3a6caa537c6 13f LazyCompile:~getConstructorName node:internal/util/inspect:567
+3a6caa53a96 d LazyCompile:~isInstanceof node:internal/util/inspect:559
+|}
+      in
+      String.split sample ~on:'\n'
+      |> List.map ~f:(fun line -> line, parse_line line)
+      |> [%sexp_of: (string * (int64 * Perf_map_location.t) option) list]
```
Xyene, line 65:
```suggestion
      |> [%sexp_of: (string * (Int64.Hex.t * Perf_map_location.t) option) list]
```

here too I guess, so that it's obvious that e.g. `3a6caa49c2e` is getting parsed correctly as `4014899239982`

cgaebel: done

Xyene (approved):
LGTM, nice work!

Issue #44, opened 2022-03-18
Create some demos


cgaebel:
Using magic-trace is a visual experience. We should create example programs and demonstrations of magic-tracing them in a few different formats for people's varying attention spans.

- a < 10s gif like [speedscope](https://github.com/jlfwong/speedscope) has, demonstrating the value magic-trace can provide
- a 2 minute video quickly walking through a short debugging session
- A text-based walkthrough where we spell out exactly what commands to run at each step

cgaebel:
I did the first and last of these. We should do the third at some point as a recorded tech talk, but I don't think that's worth keeping an issue open for.

PR #48, opened 2022-03-19
Add some example programs

cgaebel:
These programs are meant to be used with magic trace to quickly test out
whether or not it's working properly. Through varying degrees of
trickery, they can all be run via shebang.

All except the node.js example pipe bytes from /dev/zero to /dev/null.
The node.js example sends HTTP requests to itself. There's no principled
reason why it doesn't do the same thing, I just found HTTP easier deal
with in node than files.

The java demo is just called `java` instead of `demo.java` because
giving it a proper file extension breaks the shebang.

Partially addresses #44. But I don't consider it a complete solution to that
issue. These are demos, but they're not very accessible to people yet.

### DIFF
--- a/demo/demo.c
+++ b/demo/demo.c
@@ -1,0 +1,19 @@
+///usr/bin/env -S clang -gdwarf-4 "$0" -o /tmp/demo && exec /tmp/demo "$@"
+#include <unistd.h>
+#include <fcntl.h>
+
+int main(void)
+{
+    int r = open("/dev/zero", O_RDONLY);
+    int w = open("/dev/null", O_WRONLY);
+    char buf[4096] = {0};
+    for (;;)
+    {
+        ssize_t bytes_read = read(r, buf, sizeof(buf));
+        if (bytes_read < 0)
+            break;
+        if (write(w, buf, (size_t)bytes_read) < 0)
+            break;
+    }
+    return 0;
+}
--- a/demo/java
+++ b/demo/java
@@ -1,0 +1,17 @@
+#!/usr/bin/env -S java --source 11
+
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
+public class java {
+    public static void main(String[] args) throws IOException {
+        FileReader r = new FileReader("/dev/zero");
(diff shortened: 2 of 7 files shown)


demo/demo.c
```diff
@@ -0,0 +1,19 @@
+///usr/bin/env -S clang -gdwarf-4 "$0" -o /tmp/demo && exec /tmp/demo "$@"
+#include <unistd.h>
+#include <fcntl.h>
+
+int main()
```
Xyene, line 5:
```suggestion
int main(void)
```

demo/java
```diff
@@ -0,0 +1,18 @@
...
+    public static void main(String[] args) throws IOException {
+        FileReader r = new FileReader("/dev/zero");
+        FileWriter w = new FileWriter("/dev/null");
+        char[] buf = new char[4096];
+        for (;;) {
+            int bytes_read = r.read(buf);
+            w.write(buf, 0, bytes_read);
+        }
+    }
+
```
Xyene, line 17:
```suggestion
```

demo/java
```diff
@@ -0,0 +1,18 @@
...
+import java.io.FileWriter;
+import java.io.IOException;
+
+public class java {
+    public static void main(String[] args) throws IOException {
+        FileReader r = new FileReader("/dev/zero");
+        FileWriter w = new FileWriter("/dev/null");
+        char[] buf = new char[4096];
+        for (;;) {
+            int bytes_read = r.read(buf);
```
Xyene, line 13:
```suggestion
            int bytesRead = r.read(buf);
```

demo/java
```diff
@@ -0,0 +1,18 @@
...
+            w.write(buf, 0, bytes_read);
```
Xyene, line 14:
```suggestion
            w.write(buf, 0, bytesRead);
```

Xyene (commented):
Just some minor nits, otherwise LGTM

cgaebel:
I made the changes but github ICE'd when I pushed and isn't showing that fact in its GUI.

Xyene (approved):
LGTM

PR #49, opened 2022-03-20
Allow tracing all processes when root

### LINKED ISSUE #20: Process search doesn't show non-root processes when running as root
It should, because those processes are still traceable.

Xyene:
Closes #20.


cgaebel (approved):
LGTM, thanks.

PR #52, opened 2022-03-22
Refactor perf capability detection

Xyene:
This'll make future PRs cleaner. For instance, we'll be able to detect the necessary perf version for `--kcore` in https://github.com/janestreet/magic-trace/pull/21.


src/perf_tool_backend.ml
```diff
@@ -2,105 +2,166 @@ open! Core
...
-  in
-  { multi_thread; full_execution }
-;;
+let debug_perf_commands = false
 
-type recording =
-  { can_snapshot : bool
-  ; pid : Pid.t
-  }
+module Perf_capabilities = struct
```
cgaebel, line 7: What do you think about moving this to its own file, with an mli?

cgaebel (changes requested):

cgaebel (approved):

PR #59, opened 2022-03-24
Move a bunch of floating types into modules

Xyene:


cgaebel (approved):
CI failed, but otherwise LGTM.

PR #62, opened 2022-03-25
Print symbol name and hex address when using -symbol argument

### LINKED ISSUE #61: If `-symbol` is specified, the name should also be printed
We currently print the address (as an int -- it should probably be printed as a pointer). We should also print the name, in case the user might've misselected their symbol). `name @ addr` format sounds reasonable.

(Ran into this while showing someone how to use magic-trace for the first time.)

quantum5:
This makes the output much more readable. Fixes #61.

Before:
```
[Attaching to 94060412121401]
```

After:
```
[Attaching to do_read @ 0x0000558c2573b139]
```


Xyene (approved):
LGTM, thanks!

PR #66, opened 2022-03-25
Support C++ symbols output by perf

cgaebel:
C++ symbols can contain spaces. magic-trace's perf script parsing was unable to cope with those spaces and would crash. This change:

 - adds tests with a couple example C and a C++ symbol, and
 - uses a regex in the perf output parsing code to split out the symbol.


cgaebel:
I'm not totally sure why we didn't run into this before. I guess that it's either because our other C++ tests were getting symbols from the binary's symbol table instead of the output of `perf script`, or they didn't use templates.

Xyene:
Huh, I thought this was fixed by https://github.com/janestreet/magic-trace/pull/7... did it regress at some point since then?

cgaebel:
I don't know. All I know is:

- I tested this on a C++ codebase and magic-trace crashed (pointing to an invalid kind) without this patch and doesn't crash with it.
- The test I added was failing before this patch and succeeds now.

cgaebel:
Oh wait, I see the problem. #7 fixes C++ templates after the `=>`, this patch fixes templates before the `=>`.

src/perf_tool_backend.ml
```diff
@@ -172,6 +174,49 @@ module Perf_line = struct
             (exn : exn)
             ~perf_output:(line : string)]
   ;;
+
+  let%test_module _ =
+    (module struct
+      open Core
+
+      let%expect_test "c symbol" =
```
Xyene, line 182:
```suggestion
      let%expect_test "C symbol" =
```

Nit here and below

Xyene (commented):
Ah, that makes sense, that logic was added afterwards and we didn't have tests -- thanks for adding them :)

Small nit, otherwise LGTM.

src/perf_tool_backend.ml
cgaebel: Done. Also changed the C++ test to include a space in the symbol after the `=>`.

Xyene (approved):
LGTM!

PR #72, opened 2022-03-25
Make fzf an optional runtime dependency

### LINKED ISSUE #67: Provide a better error message when fzf isn't found
When fzf isn't available in the user's PATH, we crash with an ocaml exception. It's a pretty jarring experience for someone who'se never seen one before, and a terrible first impression.

Instead, let's give them a human-readable error message that invites the user to either:

- install fzf, or
- pass -pid to perf

cgaebel:
When fzf isn't available in the user's PATH, make the `-pid` argument
mandatory. When it is available, fuzzy find a PID.

Fixes #67.

Test with: `PATH= _build/default/bin/magic_trace_bin.exe attach -output /tmp/x`

Old output:

```
(monitor.ml.Error
 (Unix.Unix_error "No such file or directory" execvp
  "((prog fzf) (argv (fzf)))")
 ("Raised at Base__Result.ok_exn in file \"src/result.ml\", line 249, characters 17-26"
  "Called from Async_kernel__Deferred1.M.map.(fun) in file \"src/deferred1.ml\", line 17, characters 40-45"
  "Called from Async_kernel__Job_queue.run_jobs in file \"src/job_queue.ml\", line 167, characters 6-47"))
No pid selected
```

New output:

```
The [-pid] argument is mandatory. magic-trace could show you a fuzzy-finding selector here if "fzf" were in your PATH, but it is not.
```

### DIFF
--- a/src/trace.ml
+++ b/src/trace.ml
@@ -4,6 +4,15 @@ open! Core
 open! Async
 open! Import
 
+let supports_fzf =
+  Lazy.from_fun (fun () ->
+      let pid = Core_unix.fork_exec ~prog:"fzf" ~argv:[ "--version" ] ~use_path:true () in
+      let exit_or_signal = Core_unix.waitpid pid in
+      match Core_unix.Exit_or_signal.or_error exit_or_signal with
+      | Error _ -> false
+      | Ok () -> true)
+;;
+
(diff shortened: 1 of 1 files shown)


Xyene (approved):
LGTM, thanks!

PR #76, opened 2022-03-26
Fix fzf support detection

cgaebel:
It had two bugs, which are fixed in this change:

1) When detecting whether or not fzf is avilable, fzf was being invoked
   with argv=[--version] instead of argv=[fzf; --version]. That resulted
   it in thinking it was invoked without any arguments, which made it
   select files.

2) fzf --version was outputting to magic trace's stdout, instead of
   being swallowed.


Xyene (approved):
LGTM

PR #78, opened 2022-03-27
Add support for more perf outputs to [Perf_line.to_event] 

billduff:
Added support for more perf outputs to [Perf_line.to_event] in src/perf_tool_backend.
Simplified the code a bit along the way.

In particular, fixed a bug where any symbol with '=' in it would cause a parse error,
and added a test for this case based on a real program, as well as some more contrived
stress-tests.

I also noticed that traces from the demo Go program (at least sometimes) fail to parse,
so I added some stub code pointing out the problem.

Signed-off-by: Bill Duff <wduff11@gmail.com>

### DIFF
--- a/src/perf_tool_backend.ml
+++ b/src/perf_tool_backend.ml
@@ -113,58 +113,92 @@ module Perf_line = struct
   let report_itraces = "b"
   let report_fields = "pid,tid,time,flags,ip,addr,sym,symoff"
 
-  let kind_and_ip_re =
+  let line_re =
     Re.Posix.re
-      {|(call|return|tr strt|tr end|tr end  call|tr end  return|tr end  syscall|jmp|jcc) +([0-9a-f]+) (.*)|}
+      {|^ *([0-9]+)/([0-9]+) +([0-9]+).([0-9]+): +(call|return|tr strt|tr end|tr strt tr end|tr end  call|tr end  return|tr end  syscall|jmp|jcc) +([0-9a-f]+) (.*) => +([0-9a-f]+) (.*)$|}
     |> Re.compile
   ;;
 
-  let to_event line =
+  let symbol_and_offset_re = Re.Posix.re {|^(.*)\+(0x[0-9a-f]+)$|} |> Re.compile
+
+  let to_event line : Event.t =
     try
-      Scanf.sscanf
-        line
-        " %d/%d %d.%d: %s@=> %Lx %s@$"
-        (fun pid tid time_hi time_lo kind_and_ip addr rest ->
-          match Re.Group.all (Re.exec kind_and_ip_re kind_and_ip) with
-          | [| _; kind; ip_string; ip_rest |] ->
-            let ip =
-              try Scanf.sscanf ip_string "%Lx" (fun ip -> ip) with
-              | Scanf.Scan_failure _ | End_of_file -> 0L
-            in
-            let parse_symbol_offset str =
-              try Scanf.sscanf str "%s@+0x%x" (fun symbol offset -> symbol, offset) with
-              | Scanf.Scan_failure _ | End_of_file -> "[unknown]", 0
-            in
-            let ip_symbol, ip_offset = parse_symbol_offset ip_rest in
-            let symbol, offset = parse_symbol_offset rest in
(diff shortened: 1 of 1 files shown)


cgaebel:
I'm leaving this for Tudor to review because I helped you make it.

src/perf_tool_backend.ml
```diff
@@ -113,58 +113,90 @@ module Perf_line = struct
...
-                "Regex of expected perf output did not match." (results : string array)])
+        ; time = time_lo + (time_hi * 1_000_000_000) |> Time_ns.Span.of_int_ns
+        ; kind =
+            (match String.strip kind with
+            | "call" -> Call
+            | "return" -> Return
+            | "tr strt" -> Start
+            | "tr end" -> End None
+            | "tr strt tr end" ->
+              (* CR-someday wduff: I saw this in practice, but I'm not sure what we want to do with
```
Xyene, line 176: Could you also file a ticket with some `perf script` output around this region, so we don't forget? I weakly think this should be treated as a jump, but we should check the perf code for what makes it generate this sequence.

src/perf_tool_backend.ml
```diff
@@ -113,58 +113,90 @@ module Perf_line = struct
...
+            | "tr end  call" -> End Call
+            | "tr end  return" -> End Return
+            | "tr end  syscall" -> End Syscall
+            | "jmp" -> Jump
+            | "jcc" -> Jump
+            | kind -> raise_s [%message "unrecognized perf event" (kind : string)])
+            (* CR-someday wduff: These names make a lot more sense to me than the names that were here
+           before, but maybe I'm missing some context. We should either change the names in
+           [Event.t], or change them here, or something in between. That said, it seems best to
+           separate figuring out the names from the present improvements. *)
```
Xyene, line 188: This makes sense to me, you're not missing context.

src/perf_tool_backend.ml
Xyene: Could we collapse this as `| "tr strt  tr end" | kind` though, just to have only one error branch?

src/perf_tool_backend.ml
```diff
@@ -113,58 +113,90 @@ module Perf_line = struct
...
+        in
+        let time_hi = Int.of_string time_hi in
+        let int64_of_hex_string str =
+          try Scanf.sscanf str "%Lx" Fn.id with
+          | Scanf.Scan_failure _ | End_of_file -> 0L
+        in
+        let src_instruction_pointer = int64_of_hex_string src_instruction_pointer in
+        let dst_instruction_pointer = int64_of_hex_string dst_instruction_pointer in
+        let parse_symbol_and_offset str =
+          try Scanf.sscanf str "%s@+0x%x" (fun symbol offset -> symbol, offset) with
```
Xyene, line 155: Would you expect us to see any perf improvements by using a pre-baked regex here, over re-parsing this scan string twice per line? Not necessary to change in this PR, but I'm thinking it might be worth doing -- decoding the trace takes a long time right now.

Xyene (approved):
LGTM, thanks! Just left some comments for discussion.

src/perf_tool_backend.ml
billduff: On reflection I realized that regardless of the performance issue, this is also a bug. We'll return "[unknown]" on symbols containing "+".  Changing to a regex fixes it, so I added a test and did that. Not sure how or whether it will affect the performance because I don't have a good performance model of either Scanf or Re.

src/perf_tool_backend.ml
billduff: Sure. I left them separate so to distinguish "we don't know what to do" from "we don't recognize this at all", but that seems kinda silly given that we're not being all that careful about error messages in the first place, and either way the fix is to go in and add support for the missing case.

I changed it to your suggestion.

billduff: And opened issue #80 as well.

src/perf_tool_backend.ml
Xyene: Good catch!

Xyene (approved):
LGTM, thanks!

Issue #69, opened 2022-03-25
support non-standard characters in symbol names on command line


ahh:
I have a binary with a symbol whose c name according to e.g. readelf is like:

```
foo_bar_$5_baz$3_qux_12345
```

This fails:
```
magic-trace attach <stuff> -symbol 'foo_bar_$5_baz$3_qux_12345'
```
Surprisingly if I do:
```
magic-trace attach <stuff> -symbol foo_bar
```
I see the right symbol in my fzf, can select it, and everything works (it does not appear to have any escapes etc.)  I wonder if we're not escaping the characters in a regex? I also wonder if we should separate symbol-regex from this-is-the-symbol-i-know-it.

ahh:
(unsurprisingly, this is trivial to replicate with an ocaml anonymous function name.)

cgaebel:
How do you all feel about only supporting *exact* symbol name matches? I worry about cases like two symbols existing `symbol` and `symbol_2` and getting traces triggered off of `symbol_2` when I really want to only trace symbol.

I suppose we could offer multiple ways to select symbols, but that adds more complexity to our UI.

cgaebel:
Fixed by #79.

ahh:
It's an interesting point; I think the fzf is a really useful, dare I say
magic feature but confusion can happen.

We could special case exact match?

Issue #37, opened 2022-03-18
Clean up the CLI surface


cgaebel:
The CLI surface of magic-trace could use some polish. Here is a small list of minor complaints I'd like to address before 1.0:

- [ ] It could use a `stable v1` subcommand with a limited set of features we're comfortable committing to forever.
- [ ] We should either bundle our fork of perfetto or always point people to the web ui, instead of providing the optional `-perfetto-ui-base-directory` flag.
- [ ] Document the `decode` command and maybe deprioritize it in the command hierarchy. It's useful if another app wants to do something with perf that magic-trace doesn't support but does want to use magic-trace to visualize the result.
- [x] If `immediate-stop` mode can truly cause system crashes, let's remove it.
- [ ] There are two "trigger modes" that one can run magic-trace in: trigger when the app calls a function, or trigger when ctrl+c is hit. Let's find a way to hint that directly, maybe with separate subcommand or maybe by rejiggering the flags.
- [ ] duration_thresh references mark_start, but doesn't explain what that is
- [ ] serve should probably take the http port as an argument (probably with a default), instead of having two separate arguments
- [x] consider copying perf and outputting trace.data into the working directory, moving any existing file there into trace.data.old
- [ ] Does `full-execution` even work?
- [ ] If `-multi-thread` works, run in that mode by default and turn the flag into `-single-thread`... and maybe accept a tid?

PR #79, opened 2022-03-27
Rework trigger modes

### LINKED ISSUE #60: If `-symbol` is specified, SIGINT shouldn't capture a trace
Otherwise there's no good way to exit if you realize you've accidentally selected the wrong symbol.

(Ran into this while showing someone how to use `magic-trace` for the first time.)

cgaebel:
Replace `-symbol` with a "trigger mode" option. To quote my own help
text:

1) If you do not provide `-trigger`, magic-trace takes a snapshot
   when either it or the application under trace ends. You can Ctrl+C
   magic-trace to manually trigger a snapshot.

2) If you provide `-trigger $`, magic-trace will open up a
   fuzzy-find dialog to help you choose a symbol to trace.

3) If you provide `-trigger <FUNCTION NAME>`, magic-trace will
   snapshot when the application being traced calls the given
   function. This takes the fully-mangled name, so if you're using
   anything except C, fuzzy-find mode will probably be easier to
   use.

`-trigger .` is a shorthand for `-trigger magic_trace_stop_indicator`.

This also removes three command-line arguments that are, in my opinion,
a mix of unnecessary and probably a bad idea to expose:

- `-delay-thresh` can be replicated by an applicationg doing this
  measurement itself. It's a very cheap operation for an application
  to add, I don't think that magic-trace is adding very much here.

- `-duration-thresh` is in the same boat as `-delay-thresh`.

- `-immediate-stop` causes issues on some kernels. Let's just remove
  it for now and revisit it in a couple years. If people actually
  want their trace to end exactly when the trigger fires, we can
  postprocess the trace output.

Please play around with this! I think it's a easier to to use... but
that's just like my opinion, man.

Fixes #60 and #69.

### DIFF
--- a/core/elf.ml
+++ b/core/elf.ml
@@ -41,18 +41,10 @@ let create filename =
       let statically_mappable = is_non_pie_executable header in
       let debug = Owee_elf.find_section_body buf sections ~section_name:".debug_line" in
       Some { string; symbol; debug; base_offset; filename; statically_mappable }
-    | _, _ ->
-      Core.eprintf
-        "[Unable to find string or symbol table, will be unable to trigger on specific \
-         symbol. Was the binary built without debug info?]\n\
-         %!";
-      None
+    | _, _ -> None
   with
   | Owee_buf.Invalid_format _ ->
-    Core.eprintf
-      "[Invalid or unknown debug info format, will be unable to trigger on specific \
-       symbol.]\n\
-       %!";
+    Core.eprintf "Invalid or unknown debug info format.\n";
     None
 ;;
 
@@ -74,6 +66,33 @@ let matching_functions t symbol_re =
         | `Duplicate -> ())
       | _ -> ());
   !res
+;;
+
+let find_symbol t name =
+  let some_name = Some name in
+  with_return (fun return ->
+      Owee_elf.Symbol_table.iter t.symbol ~f:(fun symbol ->
+          if is_func symbol
+             && [%compare.equal: string option]
+                  (Owee_elf.Symbol_table.Symbol.name symbol t.string)
+                  some_name
+          then return.return (Some symbol));
+      None)
+;;
+
+let all_symbols t =
+  let res = String.Table.create () in
+  Owee_elf.Symbol_table.iter t.symbol ~f:(fun symbol ->
+      if is_func symbol
+      then (
+        match Owee_elf.Symbol_table.Symbol.name symbol t.string with
+        | None -> ()
+        | Some name ->
+          (* Duplicate symbols are possible if a symbol is in both the dynamic and static
+           symbol tables. *)
+          (match Hashtbl.add res ~key:name ~data:symbol with
+          | `Ok | `Duplicate -> ())));
+  String.Table.to_alist res
 ;;
 
 let symbol_stop_info t pid symbol =
--- a/src/trace.ml
+++ b/src/trace.ml
@@ -18,6 +18,16 @@ let supports_fzf =
         | Ok () -> true))
 ;;
 
+let create_elf ~executable ~(when_to_snapshot : When_to_snapshot.t) =
+  let elf = Elf.create executable in
+  match when_to_snapshot, elf with
+  | Application_calls_a_function _, None ->
+    Deferred.Or_error.errorf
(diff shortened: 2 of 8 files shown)


cgaebel:
Relates to, but does not entirely fix, #37.

cgaebel:
Fixes #60.

cgaebel:
Fixes #69.

Xyene:
> Fixes https://github.com/janestreet/magic-trace/issues/60.

If you put that string at the end of your commit text, it'll auto-close the issue and provide a backref :)

> `-immediate-stop` causes issues on some kernels. Let's just remove
it for now and revisit it in a couple years. If people actually
want their trace to end exactly when the trigger fires, we can
postprocess the trace output.

We should open a ticket to track this, especially the postprocessing step. I think it's necessary: I was showing someone around the tool, and their very first trace had 9 *seconds* of a locking function following the trace snapshot point... which was smushed into a pixel on the very left of the timeline. Even I was confused.

src/trace.ml
```diff
@@ -18,6 +18,16 @@ let supports_fzf =
         | Ok () -> true))
 ;;
 
+let create_elf ~executable ~(when_to_snapshot : When_to_snapshot.t) =
+  let elf = Elf.create executable in
+  match when_to_snapshot, elf with
+  | Application_calls_a_function _, None ->
+    Deferred.Or_error.errorf
+      "As far as magic-trace can tell, that executable doesn't have a symbol table. Was \
```
Xyene, line 26: I don't think this should be a fatal error condition. For instance, you can imagine not having symbols for the `java` binary, but still get mileage out of magic-trace from the perf map files. Likewise, we may not have symbols for the main binary, but have symbols for dynamically loaded libraries, and still find magic-trace useful for introspection.

Xyene: Oh, hold on, I think I misread this code...

src/trace.ml
```diff
@@ -171,65 +178,23 @@ module Make_commands (Backend : Backend_intf.S) = struct
...
-    let maybe_take_snapshot hit =
-      match opts.duration_thresh with
-      | None -> maybe_take_snapshot' hit
-      | Some duration_thresh ->
-        let _, { Breakpoint.Hit.timestamp; passed_timestamp; _ } = hit in
-        let duration = Time_ns.Span.of_int_ns (timestamp - passed_timestamp) in
-        if Time_ns.Span.( > ) duration duration_thresh then maybe_take_snapshot' hit
-    in
+    let maybe_take_snapshot' hit = take_snapshot_on_hit hit in
+    let maybe_take_snapshot hit = maybe_take_snapshot' hit in
```
Xyene, line 186: Why are these two functions?

cgaebel:
> > `-immediate-stop` causes issues on some kernels. Let's just remove
> > it for now and revisit it in a couple years. If people actually
> > want their trace to end exactly when the trigger fires, we can
> > postprocess the trace output.
> 
> We should open a ticket to track this, especially the postprocessing step. I think it's necessary: I was showing someone around the tool, and their very first trace had 9 _seconds_ of a locking function following the trace snapshot point... which was smushed into a pixel on the very left of the timeline. Even I was confused.

#81

src/trace.ml
cgaebel: No reason, I'm just bad at deleting code.

core/elf.ml
```diff
@@ -41,18 +41,10 @@ let create filename =
...
-         %!";
-      None
+    | _, _ -> None
   with
   | Owee_buf.Invalid_format _ ->
-    Core.eprintf
-      "[Invalid or unknown debug info format, will be unable to trigger on specific \
-       symbol.]\n\
-       %!";
+    Core.eprintf "Invalid or unknown debug info format.\n";
```
Xyene, line 47: Other places currently print errors/messages in `[ ... ]` (I assume to mirror `perf`). Do we want to drop that convention, or continue using it here?

Xyene (approved):
Looks good to me. I guess we'll find out if other people like the CLI changes too :)

core/elf.ml
cgaebel: I kind of want to drop it, and also maaaaaaybe silence the perf output (I like seeing the trace size, but it's mostly noise).

PR #85, opened 2022-03-27
Move README image assets into docs/assets

Xyene:
So that they don't disappear when GitHub decides they should disappear.


cgaebel:
LGTM, thanks!

cgaebel (approved):

Issue #71, opened 2022-03-25
CI artifacts should be built with flambda


Xyene:
The decoding stage gets a fair bit faster if they are.

cgaebel:
Fixed by #87

PR #87, opened 2022-03-27
Build binary with flambda on CI

quantum5:
This PR makes the normal build artifact use flambda. To avoid accidentally regressing, the cache key is changed so that the previous non-flambda cache is not used.

Refs #71.


cgaebel:
LGTM, thanks a ton!

PR #90, opened 2022-03-28
Default `-output` to `trace.ftf`

Xyene:
This better mirrors what `perf` does with `perf.data`, so should be
familiar to users.


cgaebel:
Feel free to leave this to a new feature, but you should do the perf.data.old trick, too.

cgaebel (changes requested):
You should replicate the perf.data.old trick, too.

Xyene:
Good idea, done.

cgaebel (approved):

PR #95, opened 2022-03-28
Represent pointers as pointers in the Fuchsia trace

### LINKED ISSUE #46: Represent pointers as pointers in the Fuchsia trace
We represent them as signed `int63`s right now.

Xyene:
Closes #46.


cgaebel (approved):
LGTM, thanks!

PR #96, opened 2022-03-29
Vendor `owee` and `ocaml-probes`

Xyene:
We'll need it for #2 and proper OCaml exception handling for the near
future.


cgaebel (approved):
LGTM :(

gretay-js:
We don't need to vendor `ocaml-probes`. Can the two functions (unrelated to probes) that "magic-trace trace" uses just be copied here? They would need to be changed anyway, to avoid the race condition on tiny programs that @cgaebel mentioned elsewhere.

Xyene:
`ocaml-probes` is vendored here just because it pulls in a different `owee` when `opam install --deps-only` is run, which conflicts with the vendored `owee`.

I intend to fix `trace` sometime, at which point we can drop `ocaml-probes` as a dependency anyway, as you say.

PR #97, opened 2022-03-29
Proper OCaml exception tracking

Xyene:


src/trace.ml
```diff
@@ -83,6 +86,79 @@ module Make_commands (Backend : Backend_intf.S) = struct
...
+    let print_traps t =
+      print "enter_traps:" t.enter_traps;
+      print "push_traps:" t.push_traps;
+      print "pop_traps:" t.pop_traps
+    ;;
+
+    let make elf =
+      let path = Elf.filename elf in
+      Core.print_s [%message "file" (path : string)];
+      let buffer = Owee_buf.map_binary path in
```
gretay-js, line 133: This is duplicating the work of `map_binary` and `read_elf` from `core/elf.ml`

src/trace_writer.ml
```diff
@@ -41,19 +41,145 @@ module Thread_info = struct
       mutable callstack : Callstack.t
     ; inactive_callstacks : Callstack.t Stack.t
     ; mutable last_decode_error_time : Time_ns.Span.t
-    ; (* Currently keeping track of the number of frames to unwind during an exception by
```
gretay-js, line 44: We should keep this workflow and use it when the binary is compiled without the new elf notes.

src/trace.ml
Xyene: Yeah, this is terrible code that's not quite ready for review, was focusing on getting it working :)

Xyene:
Superseded by https://github.com/janestreet/magic-trace/pull/97.

PR #98, opened 2022-03-30
Add an go example that actually works

cgaebel:
Go code is still a little broken because it has custom control flow when it switches stacks. We'll fix that one day, but not today.

Screen Shot 2022-03-29 at 10 54 26 PM


Xyene (approved):
LGTM!

PR #99, opened 2022-03-30
Build binary with flambda on CI (try 2)

quantum5:
Last time, this failed due to additional complexity introduced by ocaml/setup-ocaml@v2. This complexity is supposed to make our lives better, but it doesn't: it even interferes with our cache, forcing you to use theirs. Except... their cache explicitly doesn't cache our dependencies, so it's still slow. Adding insult to injury, it's also installing more packages from apt, which can't be cached.

Since it's such a pain, we will just do what the v1 did with raw commands, except installing flambda.

Example run without cache: https://github.com/quantum5/magic-trace/runs/5750942329
Example run with cache: https://github.com/quantum5/magic-trace/runs/5751309514

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,16 +21,24 @@ jobs:
       - uses: actions/cache@v2
         with:
           path: ~/.opam
-          key: ${{ matrix.os }}-opam-${{ matrix.ocaml-version }}
-          restore-keys: |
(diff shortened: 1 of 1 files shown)


cgaebel (approved):
LGTM. I'm not sure if you intentionally snuck in that musl-tools dependency, but I approve either way.

quantum5:
For some reason the v1 of setup-ocaml installs those apt packages so I just copied them.

PR #101, opened 2022-03-30
Clean up the CLI surface

### LINKED ISSUE #37: Clean up the CLI surface

cgaebel:
Document each argument and refactor the commands a little bit to support
the common pattern [-p $PID].

Fixes #37.

Help text before:

```
Attach to a process and record it until Ctrl-C, then convert the results to a viewable Fuchsia trace.

  magic_trace_bin.exe attach

=== flags ===

  [-executable-override FILE]
                             . executable to extract information from, default
                               is to use the first part of COMMAND
  [-full-execution]          . record full program execution
  [-http-port PORT]          . http server port
  [-multi-snapshot]          . allow taking multiple snapshots if possible
  [-multi-thread]            . record multiple threads
  [-output FILE]             . output file name (default: 'trace.ftf')
  [-perfetto-ui-base-directory PATH]
                             . path to Perfetto in filesystem
  [-pid PID]                 . Process to attach to. Required if you don't have
                               the "fzf" application available in your PATH.
  [-record-dir DIR]          . create this directory if necessary and put raw
                               trace data in it
  [-serve]                   . also serve the trace
  [-trace-include-kernel]    . trace userspace and kernel
  [-trace-kernel-only]       . trace kernel only
  [-trace-userspace]         . trace userspace only
  [-trigger SYMBOL]          . Decides when magic-trace takes a trace. There are
                               three trigger modes to choose from:
                               1) If you do not provide [-trigger], magic-trace
                               takes a snapshot when either it or the
                               application under trace ends. You can Ctrl+C
                               magic-trace to manually trigger a snapshot.
                               2) If you provide [-trigger ?], magic-trace will
                               open up a fuzzy-find dialog to help you choose a
                               symbol to trace.
                               3) If you provide [-trigger <FUNCTION NAME>],
                               magic-trace will snapshot when the application
                               being traced calls the given function. This takes
                               the fully-mangled name, so if you're using
                               anything except C, fuzzy-find mode will probably
                               be easier to use. [-trigger .] is a shorthand for
                               [-trigger magic_trace_stop_indicator].
  [-verbose]                 . print decoded events
  [-help], -?                . print this help text and exit
```

Help text after:

```
Traces a running process.

  magic_trace_bin.exe attach

=== flags ===

  [-full-execution]          . Record a program's full execution instead of
                               using a snapshot ring buffer.
                               Warning: The trace grows at a rate of hundreds of
                               megabytes per second. Traces larger than 100 MiB
                               are likely to crash the trace viewer.
  [-multi-snapshot]          . Take a snapshot every time the trigger is hit,
                               instead of only the first time. This flag has two
                               caveats:
                               (1) There's an ~8us performance hit every time
                               the trigger symbol is hit. If snapshots trigger
                               frequently, your application's performance may be
                               materially impacted.
                               (2) Each snapshot linearly increases the size of
                               the trace file. Large trace files may crash the
                               trace viewer.
  [-multi-thread]            . Records every thread of an executable, instead of
                               only the thread whose TID is equal to the
                               process' PID.
                               Warning: this flag decreases the trace's lookback
                               period because the kernel divides snapshot buffer
                               resources equally across all threads.
  [-output FILE]             . Where to output the trace. (default: 'trace.ftf')
  [-pid PID]                 . Process to attach to. Required if you don't have
                               the "fzf" application available in your PATH.
  [-trace-include-kernel]    . Trace userspace and the kernel. Requires root.
  [-trace-kernel-only]       . Trace the kernel and do not trace userspace.
                               Requires root.
  [-trigger _]               . Selects when magic-trace takes a trace. There are
                               three trigger modes to choose from:
                               (1) If you do not provide [-trigger], magic-trace
                               takes a snapshot when you Ctrl+C it.
                               (2) If you provide [-trigger ?], magic-trace will
                               open up a fuzzy-find dialog to help you choose a
                               symbol to trace. Fails if you don't have the
                               "fzf" binary in your PATH.
                               (3) If you provide [-trigger <FUNCTION NAME>],
                               magic-trace will snapshot when the application
                               being traced calls the given function. This takes
                               the fully-mangled name, so if you're using
                               anything except C, fuzzy-find mode will probably
                               be easier to use. [-trigger .] is a shorthand for
                               [-trigger magic_trace_stop_indicator].
                               (*) Regardless of trigger mode, magic-trace will
                               always snapshot when the application terminates
                               if it has not yet triggered for any other reason.
  [-verbose]                 . print decoded events
  [-working-directory DIR]   . Where to store intermediate files (including raw
                               perf.data files). If not provided, magic-trace
                               stores them in a subdirectory of $TMPDIR and
                               deletes them when it's done. If provided, files
                               will be stored in the given directory, creating
                               the directory if necessary, and magic-trace will
                               not delete the directory when it's done.
  [-help], -?                . print this help text and exit
```

### DIFF
--- a/src/perf_tool_backend.ml
+++ b/src/perf_tool_backend.ml
@@ -12,9 +12,22 @@ module Record_opts = struct
 
   let param =
     let%map_open.Command multi_thread =
-      flag "-multi-thread" no_arg ~doc:"record multiple threads"
+      flag
+        "-multi-thread"
+        no_arg
+        ~doc:
+          "Records every thread of an executable, instead of only the thread whose TID \
+           is equal to the process' PID.\n\
+           Warning: this flag decreases the trace's lookback period because the kernel \
+           divides snapshot buffer resources equally across all threads."
     and full_execution =
-      flag "-full-execution" no_arg ~doc:"record full program execution"
+      flag
+        "-full-execution"
+        no_arg
+        ~doc:
+          "Record a program's full execution instead of using a snapshot ring buffer.\n\
+           Warning: The trace grows at a rate of hundreds of megabytes per second. \
+           Traces larger than 100 MiB are likely to crash the trace viewer."
     in
     { multi_thread; full_execution }
   ;;
--- a/src/trace.ml
+++ b/src/trace.ml
@@ -90,7 +90,7 @@ module Make_commands (Backend : Backend_intf.S) = struct
       { Decode_opts.output_config; decode_opts; verbose }
     =
     Core.eprintf "[Decoding, this may take 30s or so...]\n%!";
-    Tracing_tool_output.write_and_view output_config ~f:(fun w ->
+    Tracing_tool_output.write_and_maybe_view output_config ~f:(fun w ->
         let open Deferred.Or_error.Let_syntax in
         let trace_writer = Tracing.Trace.Expert.create ~base_time:None w in
         let hits =
@@ -245,9 +245,9 @@ module Make_commands (Backend : Backend_intf.S) = struct
     Backend.Recording.finish_recording recording
   ;;
 
-  let run_and_record ~elf ~command record_opts =
+  let run_and_record ~elf ~command_with_args record_opts =
     let open Deferred.Or_error.Let_syntax in
-    let argv = Array.of_list command in
+    let argv = Array.of_list command_with_args in
     let pid = Probes_lib.Raw_ptrace.start ~argv |> Pid.of_int in
     let%bind attachment = attach ~elf record_opts pid in
     Probes_lib.Raw_ptrace.detach (Pid.to_int pid);
@@ -272,31 +272,34 @@ module Make_commands (Backend : Backend_intf.S) = struct
   let record_dir_flag mode =
     let open Command.Param in
     flag
-      "-record-dir"
+      "-working-directory"
       (mode Filename_unix.arg_type)
-      ~doc:"DIR create this directory if necessary and put raw trace data in it"
+      ~doc:
+        "DIR Where to store intermediate files (including raw perf.data files). If not \
+         provided, magic-trace stores them in a subdirectory of $TMPDIR and deletes them \
+         when it's done. If provided, files will be stored in the given directory, \
+         creating the directory if necessary, and magic-trace will not delete the \
+         directory when it's done."
   ;;
 
   let record_flags =
     let%map_open.Command record_dir = record_dir_flag optional
-    and executable_override =
-      flag
-        "-executable-override"
-        (optional Filename_unix.arg_type)
-        ~doc:
-          "FILE executable to extract information from, default is to use the first part \
-           of COMMAND"
     and when_to_snapshot = When_to_snapshot.param
     and multi_snapshot =
-      flag "-multi-snapshot" no_arg ~doc:"allow taking multiple snapshots if possible"
+      flag
+        "-multi-snapshot"
+        no_arg
+        ~doc:
+          "Take a snapshot every time the trigger is hit, instead of only the first \
+           time. This flag has two caveats:\n\
+           (1) There's an ~8us performance hit every time the trigger symbol is hit. If \
+           snapshots trigger frequently, your application's performance may be \
+           materially impacted.\n\
+           (2) Each snapshot linearly increases the size of the trace file. Large trace \
+           files may crash the trace viewer."
     and trace_mode = Trace_mode.param
     and backend_opts = Backend.Record_opts.param in
-    fun ~default_executable ~f ->
-      let executable =
-        match executable_override with
-        | Some path -> path
-        | None -> default_executable ()
-      in
+    fun ~executable ~f ->
       let record_dir, cleanup =
         match record_dir with
         | Some dir ->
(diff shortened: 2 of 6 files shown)


src/perf_tool_backend.ml
```diff
@@ -12,9 +12,22 @@ module Record_opts = struct
...
+           divides snapshot buffer resources equally across all threads."
     and full_execution =
-      flag "-full-execution" no_arg ~doc:"record full program execution"
+      flag
+        "-full-execution"
+        no_arg
+        ~doc:
+          "Record a program's full execution instead of using a snapshot ring buffer.\n\
+           Warning: The trace grows at a rate of hundreds of megabytes per second. \
+           Traces larger than 100 MB are likely to crash the trace viewer."
```
Xyene, line 30:
```suggestion
           Traces larger than 100 MiB are likely to crash the trace viewer."
```

src/trace.ml
```diff
@@ -272,23 +272,46 @@ module Make_commands (Backend : Backend_intf.S) = struct
   let record_dir_flag mode =
     let open Command.Param in
     flag
-      "-record-dir"
+      "-temporary-data-directory"
```
Xyene, line 275: Don't think `temporary` is quite the right term here, given this directory explicitly doesn't get cleaned up.

src/trace.ml
```diff
@@ -272,23 +272,46 @@ module Make_commands (Backend : Backend_intf.S) = struct
...
       (mode Filename_unix.arg_type)
-      ~doc:"DIR create this directory if necessary and put raw trace data in it"
+      ~doc:
+        "DIR Where to store temporary files (including raw perf.data files). If not \
+         provided, magic-trace stores temporary files in a subdirectory of /tmp and \
```
Xyene, line 279: Is this actually `/tmp`, or is it `$TMPDIR`?

src/trace.ml
```diff
@@ -272,23 +272,46 @@ module Make_commands (Backend : Backend_intf.S) = struct
...
+         deletes them when it's done. If provided, files will be stored in the given \
+         directory, creating the directory if necessary, and magic-trace will not delete \
+         the directory when it's done."
   ;;
 
+  (*
```
Xyene, line 285: Should we delete this code?

src/trace.ml
```diff
@@ -272,23 +272,46 @@ module Make_commands (Backend : Backend_intf.S) = struct
...
+        | None -> Filename_unix.temp_dir "magic_trace" "", true
+      in
+  *)
+
   let record_flags =
     let%map_open.Command record_dir = record_dir_flag optional
     and executable_override =
       flag
-        "-executable-override"
+        "-override-executable"
```
Xyene, line 299: I don't feel strongly about this, but I don't know where this flag would be useful.

src/tracing_tool_output.ml
```diff
@@ -2,21 +2,37 @@ open! Core
...
+    | Enabled of enabled
+
   let param =
     let%map_open.Command port =
-      flag "http-port" (optional_with_default 8080 int) ~doc:"PORT http server port"
-    and perfetto_ui_base_directory =
+      let default = 8080 in
       flag
-        "perfetto-ui-base-directory"
+        "serve-port"
```
Xyene, line 18: We should consider hiding `-serve` behind whether `$PERFETTO_DIR` is set, or something. Also, it should probably take `optional_with_default 8080 int` itself, so we have one less flag.

src/perf_tool_backend.ml
```diff
@@ -12,9 +12,22 @@ module Record_opts = struct
...
     let%map_open.Command multi_thread =
-      flag "-multi-thread" no_arg ~doc:"record multiple threads"
+      flag
+        "-multi-thread"
+        no_arg
+        ~doc:
+          "Records every thread of an executable, instead of only the thread whose TID \
+           is equal to the process' PID.\n\
+           Warning: this flag decreases the trace's lookback period because the kernel \
+           divides snapshot buffer resources equally across all threads."
```
Xyene, line 22: I know I'm the one that said this, but I'm only 70% on this specifically. I know the resources are divided, just not exactly how.

Xyene (approved):
LGTM!

PR #107, opened 2022-03-30
Build debian package with flambda

quantum5:
Example run: https://github.com/quantum5/magic-trace/runs/5762575531


cgaebel:
Nice. Thank you!

PR #108, opened 2022-03-31
Build musl statically linked executable on CI

### LINKED ISSUE #64: Build static binaries in CI
This would require:

* building against musl
* possibly different `dune` targets

...but would allow the attached binaries to be run on anything, not just Ubuntu 20.04-glibc-equivalent systems.

quantum5:
This commit adds a dune profile called `static`, which forces dune to use static linking by passing `-cclib -static`.

It also creates the opam switch with `ocaml-option-musl` and installs a bunch of musl-compatible kernel headers from `sabotage-linux` on CI so that `core_unix` and the `#include <include/ptrace.h>` in `lib/ocaml-probes` compile.

Example run: https://github.com/quantum5/magic-trace/runs/5764578268
The resulting binaries are tested to run on a `xenial` machine.

Closes #64.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,7 +21,14 @@ jobs:
       - uses: actions/cache@v2
         with:
           path: ~/.opam
-          key: ${{ matrix.os }}-opam-${{ matrix.ocaml-version }}-flambda-v2
+          key: ${{ matrix.os }}-opam-${{ matrix.ocaml-version }}-flambda-musl
+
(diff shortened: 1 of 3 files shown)


Xyene (approved):
LGTM, thanks!

cgaebel:
This is awesome, thanks for figuring it out! I was getting nowhere with it earlier today.

```
$ ldd /tmp/magic-trace
	not a dynamic executable
```

It's beautiful.

PR #109, opened 2022-03-31
Run strip and upx on the executable built by CI

quantum5:
This reduces the executable size from ~24 MB to ~3.5 MB, at the cost of increasing the program startup time by 26 ms on my Ryzen 9 5950X.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,6 +30,12 @@ jobs:
             tar -xz -C musl-kernel --strip-components=1
           echo "C_INCLUDE_PATH=$(pwd)/musl-kernel/x86/include" >> "$GITHUB_ENV"
 
+      - name: Install upx
+        run: |
+          # The version that comes with focal is broken for musl binaries, pulling 3.96 from hirsute.
+          wget http://azure.archive.ubuntu.com/ubuntu/pool/universe/u/upx-ucl/upx-ucl_3.96-2_amd64.deb
+          sudo dpkg -i upx-ucl_3.96-2_amd64.deb
+
       - name: Use OCaml ${{ matrix.ocaml-version }}
         run: |
           sudo apt install bubblewrap musl-tools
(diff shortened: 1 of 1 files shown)


cgaebel:
I'm not sure we want to run strip. Having debug symbols makes it possible for users to trace magic-trace, which seems like a not-unlikely thing our users may try. I want them to have a good experience when they do. What's the binary size if you *don't* call strip?

I'm not sure we want upx packing. Historically it's caused virus scanners to alert, although I don't know if that's still true. I'll do more research on this front and get back to you.

cgaebel:
VirusTotal like this: https://www.virustotal.com/gui/file/3bac04231e169e35103193faed58476a2a1529bdc26753da4d6dac7630660a28. I still need to run this by our security people at work.

cgaebel:
|     | noupx  | upx |
| --- | ------------- | ------------- |
| <b>nostrip</b> | 23.8M  | 5.6M  |
| <b>strip</b>    | 18M  | 4.4M  |

Based on those results, I think I prefer unstripped UPX to stripped UPX.

I used `upx --ultra-brute`, which took ~3m to pack on my machine. Do you think that'd be reasonable to add to our overall CI latency @quantum5 @Xyene?

Xyene:
GitHub action runners are very slow, so I assume `upx --ultra-brute` would take a while longer than ~3m. I'd be a little sad to have all commits take that long, but we could probably do regular builds with `-9`, and tagged builds with `--ultra-brute`, what do you think?

Xyene:
Also, I suspect (but haven't tried) that you couldn't magic-trace upx'd non-stripped magic-trace, just because the symbol table would be compressed (unless it appears as decompressed in `/proc/$pid/exe`).

cgaebel:
If that’s easy to set up, sounds good to me.

cgaebel:
If you can’t magic trace the unstripped upx binary, let’s strip it.

quantum5:
I discovered that `--lzma` seems much faster and produces an 2.6 MB binary after stripping, albeit at the cost of taking 120 ms longer to start compared to uncompressed.

Meanwhile, the `--ultra-brute` binary produced does not actually work on my system, which probably suggests that we shouldn't use it.

cgaebel:
150ms is getting to be a bit much. `upx -9` without symbol table sounds fine to me if you confirm that upx alone can't be magic-traced usefully, and pending me asking our security people about this (ETA ~1 week, I'm on vacation).

cgaebel:
Cybersecurity signed off on this. Merging it.

cgaebel:
Cybersecurity signed off on this. Merging it.

PR #110, opened 2022-03-31
Miscellaneous fixes

Xyene:
Here's a batch of fixes that didn't deserve their own PRs.


Xyene:
I tacked on two more small cleanups to this chain.

I attempted to split up `perf_utils.h` into a C file, but that then required me to change the dunefile to add `perf_utils` as a `foreign_stub` in order for it to link... which felt worse than the small header, since they're not really stubs.  I can commit those changes if you think that's still net positive.

cgaebel (approved):
LGTM, thanks!

PR #114, opened 2022-04-01
Crash with an error if perf fails

### LINKED ISSUE #34: Handle `perf` failing to start more gracefully
If `perf` fails to start for any reason, magic-trace hangs and does not propagate the error up. An easy way to test this is to intentionally mess up the flags to `perf record`, or unintentionally by running on a too-old `perf` version.

### LINKED ISSUE #38: Fail gracefully
When magic-trace first starts up, it should detect conditions that we know will never work and bail out early with a good error message when they're hit. [This wiki page ](https://github.com/janestreet/magic-trace/wiki/Supported-platforms,-programming-languages,-and-runtimes) contains a list of constraints we know about. Let's automate as much as we can.

cgaebel:
magic-trace invokes perf twice: once to record a process and once to
turn that record into something magic-trace understands. Both of
those invocations can fail, and we weren't handling that well.

- If `perf record` failed, magic-trace would hang and print nothing
if the error message was long enough.

- If `perf script` failed, magic-trace would just think that there
are no events and not print any message about it.

After applying this change:

- `perf` errors are always forwarded to stderr, and
- if perf returns non-zero, magic-trace returns non-zero.

Testing: Messed up the arguments to the perf calls, one at a time,
and saw that a proper error was printed and that magic-trace
returned non-zero.

Fixes #34 and fixes #38.

### DIFF
--- a/core/backend_intf.ml
+++ b/core/backend_intf.ml
@@ -35,5 +35,5 @@ module type S = sig
     :  Decode_opts.t
     -> record_dir:string
     -> perf_map:Perf_map.t option
-    -> Event.t Pipe.Reader.t Deferred.Or_error.t
+    -> Decode_result.t Deferred.Or_error.t
 end
--- a/core/decode_result.ml
+++ b/core/decode_result.ml
@@ -1,0 +1,7 @@
+open Core
+open Async
+
+type t =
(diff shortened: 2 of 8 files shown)


Xyene (approved):
LGTM, great job!

PR #116, opened 2022-04-01
Gracefully handle duplicates in perf map files

cgaebel:
News to me: perf map files can contain the same address multiple times.
When that happens, take the one that appears later in the file.

I saw this happen with V8. What I think is going on is that it's reusing
address space that old code was JITed into for newly JITed code.


Xyene (approved):
Fun, LGTM

PR #117, opened 2022-04-02
Permit printing perf commands from the CLI

cgaebel:
This ability was previously configured with a global variable in the
code that requires a rebuild to switch.

I also rework how "verbose" stuff works in magic-trace. Instead of
a single "-v" flag and maybe a "-vv" flag, I opted instead to give the
two verbose flags clear names describing what they do, but start them
with -z- so they sort to the bottom of the argument list.

Also, hide debug commands behind an environment variable to avoid
polluting `-help` with flags that most users don't need.

### DIFF
--- a/core/backend_intf.ml
+++ b/core/backend_intf.ml
@@ -16,6 +16,7 @@ module type S = sig
 
     val attach_and_record
       :  Record_opts.t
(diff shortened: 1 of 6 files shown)


Xyene (approved):
LGTM, thanks!

PR #118, opened 2022-04-03
Make the snapshot buffer size configurable

### LINKED ISSUE #33: Support increasing the snapshot buffer size
The snapshot buffer can be ~arbitrarily sized, and is passed as the argument to `--snapshot` in `perf`. Larger buffers would allow for more data to be captured and displayed.

Ref https://man7.org/linux/man-pages/man1/perf-intel-pt.1.html:

       To select snapshot mode a new option has been added:

           -S

       Optionally it can be followed by the snapshot size e.g.

           -S0x100000

       The default snapshot size is the auxtrace mmap size. If neither
       auxtrace mmap size nor snapshot size is specified, then the
       default is 4MiB for privileged users (or if
       /proc/sys/kernel/perf_event_paranoid < 0), 128KiB for
       unprivileged users. If an unprivileged user does not specify mmap
       pages, the mmap pages will be reduced as described in the new
       auxtrace mmap size option section below.

       The snapshot size is displayed if the option -vv is used e.g.

           Intel PT snapshot size: %zu

cgaebel:
Provide a command-line option to tune perf's snapshot buffer size.
This comes with a ton of caveats and I'll have a wiki page up soon
with what it entails.

Testing: I tried a variety of snapshot sizes both with and without
perf_event_paranoid = -1. Warnings and errors were passed upwards
from perf as expected.

Fixes #33

### DIFF
--- a/src/perf_tool_backend.ml
+++ b/src/perf_tool_backend.ml
@@ -11,6 +11,7 @@ module Record_opts = struct
   type t =
     { multi_thread : bool
     ; full_execution : bool
+    ; snapshot_size : Pow2_pages.t option
     }
 
   let param =
@@ -29,10 +30,17 @@ module Record_opts = struct
         no_arg
         ~doc:
           "Record a program's full execution instead of using a snapshot ring buffer.\n\
-           Warning: The trace grows at a rate of hundreds of megabytes per second. \
-           Traces larger than 100 MiB are likely to crash the trace viewer."
+           Warning: The trace grows at a rate of hundreds of megabytes per second. The \
+           trace viewer may fail to load traces larger than 100M."
+    and snapshot_size =
+      Pow2_pages.optional_flag
+        "-snapshot-size"
+        ~doc:
+          " Tunes the amount of data captured in a trace. By default, 4M for root or if \
+           perf_event_paranoid = -1, 256K otherwise. The trace viewer may fail to load \
+           large traces."
     in
-    { multi_thread; full_execution }
+    { multi_thread; full_execution; snapshot_size }
   ;;
 end
 
@@ -54,7 +62,7 @@ module Recording = struct
   ;;
 
   let attach_and_record
-      { Record_opts.multi_thread; full_execution }
+      { Record_opts.multi_thread; full_execution; snapshot_size }
       ~debug_print_perf_commands
       ~(trace_mode : Trace_mode.t)
       ~record_dir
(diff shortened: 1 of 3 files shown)


src/perf_tool_backend.ml
```diff
@@ -33,8 +34,15 @@ module Record_opts = struct
           "Record a program's full execution instead of using a snapshot ring buffer.\n\
            Warning: The trace grows at a rate of hundreds of megabytes per second. \
            Traces larger than 100 MiB are likely to crash the trace viewer."
+    and snapshot_size =
+      Pow2_bytes.optional_flag
+        "-snapshot-size"
+        ~doc:
+          " Tunes the amount of data captured in a trace. By default, 4M for root or if \
```
Xyene, line 41: Nit: we use mebibyte units above, let's be consistent one way or another (weak preference for mebibytes)

Xyene: Nit: "cause the trace viewer to hit your browser's per-tab memory limit and fail to load the trace" or something, it doesn't *really* crash.

src/perf_tool_backend.ml
```diff
@@ -106,12 +114,18 @@ module Recording = struct
            %!";
         []
     in
+    let snapshot_size_arg =
```
Xyene, line 117: Nit: we call these `opt` elsewhere in this fn, let's be consistent one way or another. No preference for arg vs opt.

src/pow2_bytes.ml
```diff
@@ -0,0 +1,108 @@
+open Core
```
Xyene, line 1:
```suggestion
open! Core
```

src/perf_tool_backend.ml
Xyene: I don't know if `Bytes_unit` can take -bibyte, though...

Xyene (commented):
Just a couple of nits, otherwise this mostly looks good.

src/perf_tool_backend.ml
cgaebel: `Byte_units` can't, so I'm going to use its conventions universally (10M, 50G, etc.)

cgaebel: > Nit: "cause the trace viewer to hit your browser's per-tab memory limit and fail to load the trace" or something, it doesn't _really_ crash.

I mean... as far the user is concerned, it does. I don't feel like being more specific is particularly helpful here, and the help text is already to long imo.

cgaebel: shannon suggested "fail to load" instead of "crash"

Xyene (approved):
Looks good!

PR #119, opened 2022-04-03
Add more structure to Event.t

cgaebel:
Apply Bill's suggested renaming for `Event.t`s fields, and propagate the
new names downstream.

Also, fixed a bug where the `jmp ()` test was testing `Call`, not
`Jump`.


Xyene (approved):
Looks cleaner, thanks

PR #126, opened 2022-04-07
In help text, provide short links

cgaebel:
We have a proper domain with 301 redirects now. This uses short wiki
links to leave out detail from help text that not all people will
care about. Users that care about the full explaination can follow a
link.


Xyene (approved):
LGTM

Xyene:
Though maybe 'more info' reads a little better?

PR #130, opened 2022-04-08
Ensure `perf record` process exits when `magic-trace` does

Xyene:
Otherwise, under certain circumstances (well-timed ^C, magic-trace
crash), `perf record` remains running on the machine. This is
problematic because there can only be one IPT consumer at a time, so
getting into this state requires realizing there is still a `perf`
process running, and killing it manually.

Ideally we'd do this for the `perf script` invocation too, but I've
punted that as it would require manually doing some `Process` machinery
ourselves, and at least `perf script` is time-bounded, unlike `perf
record`.


cgaebel (approved):
LGTM

PR #131, opened 2022-04-08
Add well-known aliases `-p`/`-o` for `-pid`/`-output`

Xyene:
These should feel more familiar for anyone coming over from `perf`.

### DIFF
--- a/src/trace.ml
+++ b/src/trace.ml
@@ -449,6 +449,7 @@ module Make_commands (Backend : Backend_intf.S) = struct
          flag
            "-pid"
            (optional int)
+           ~aliases:[ "-p" ]
            ~doc:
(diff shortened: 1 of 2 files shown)


cgaebel:
Didn't we already support that? I use -p all the time. Or are you just doing this to add it to the help text?

Xyene:
It works right now by semi-accident, because we (currently) have no other flag starting with `o` or `p`. This way we can both have it in the help text + ensure it remains functional in the future.

cgaebel:
I have no problems with this change, but it definitely doesn't work by accident. It was an explicit constraint I was optimizing for when I restructured magic-trace's CLI.

cgaebel (approved):

PR #132, opened 2022-04-08
Allow sharing traces via `MAGIC_TRACE_SHARE_COMMAND`

Xyene:
Tested with `MAGIC_TRACE_SHARE_COMMAND=/usr/bin/echo`.

Also fixed a bug where we were unconditionally serving the UI if the
Perfetto base dir was specified. We lose port configurability, but it's
not a big deal and we can add it back as a separate flag if we want to.


Xyene:
I'm not sure which I like more:

- a `-serve-port` flag, or
- starting from 8080 and trying increasing port numbers until we find one we can bind to

cgaebel:
My heart says (2, but start at 1024), but my brain says (1). S'up to you.

src/tracing_tool_output.ml
```diff
@@ -2,30 +2,21 @@ open! Core
...
-      in
-      Enabled { port; perfetto_ui_base_directory }
+  let maybe_param =
+    Option.map Env_vars.perfetto_ui_base_directory ~f:(fun perfetto_ui_base_directory ->
+        let port = 8080 in
+        let%map_open.Command serve =
+          flag
+            "serve"
+            no_arg
+            ~doc:[%string " Serves the magic-trace UI on port %{port#Int})"]
```
cgaebel, line 17: extra paren at the end

src/tracing_tool_output.ml
```diff
@@ -110,8 +101,35 @@ module Serve = struct
...
+  ;;
+
+  let share_trace_file { share_command_filename } ~store_path =
+    let%map.Deferred.Or_error output =
+      Process.run
+        ~prog:share_command_filename
+        ~args:[ Filename_unix.realpath store_path ]
+        ()
+    in
+    Core.printf "%s%!" output
```
cgaebel, line 120: This won't show output until the command ends, and it might also swallow stderr? I'd use the cruddy process APIs to transfer stdout/stderr of the subprocess to magic-trace's.

src/tracing_tool_output.ml
```diff
@@ -110,8 +101,35 @@ module Serve = struct
   ;;
 end
 
+module Share = struct
+  type t = { share_command_filename : string }
+
+  let maybe_param =
+    Option.map Env_vars.share_command_filename ~f:(fun share_command_filename ->
+        let%map_open.Command share = flag "share" no_arg ~doc:"Share trace." in
```
cgaebel, line 109: That help text could say more. If you don't want it to be too long, I'm happy with the docs just being `magic-trace.org/w/share` and then we explain how sharing works on the wiki somewhere.

Serve could arguably work the same way.

cgaebel (changes requested):

Issue #81, opened 2022-03-27
Filter events from after the stop indicator


cgaebel:
Magic-trace doesn't stop recording exactly when the stop indicator is hit. That can be quite confusing, especially if you get unlucky and the stop event is nowhere near the right hand side of the trace.

To fix this, let's filter out events from the trace that happen after the stop indicator returns.

It's conceivable that people will want the current behavior behind a flag, but I weakly think it's not worth the extra UI complexity.

cgaebel:
Looks like this could be useful: search `man perf-record` for "tracestop".

Xyene:
Ref also `-immediate-stop` from https://github.com/janestreet/magic-trace/pull/79.

PR #133, opened 2022-04-08
Trim excess events off trace

Xyene:
Refs https://github.com/janestreet/magic-trace/issues/81

Don't know if we want it...


Xyene:
test comment

PR #135, opened 2022-04-08
Snapshot on exit, not on signal

### LINKED ISSUE #113: `trace` is race-y with process startup
If a process takes less than ~500ms to execute, then `trace` will produce an empty tracefile. We should synchronize process startup with `perf`, probably through some `SIGSTOP`/`SIGCONT` dance.

cgaebel:
This change makes perf snapshot on exit. For simplicity, we no longer
send perf SIGUSR2, we just send it SIGINT and ask it to snapshot on
exit.

When magic-tracing a small program with `magic-trace run $PROG`, users
would see an empty trace. We originally thought this was because
magic-trace wasn't pausing the process on startup, but that was actually
working just fine. The real reason it failed is because perf wasn't
taking a snapshot at exit -- you need to pass `--snapshot=e` for that.

Testing: Got a full trace of a C hello world program with `magic-trace
run`.
Also checked that `magic-trace attach` still works on `demo.c`.
The trace of a C hello world is really cool!

Fixes #113. We were wrong about the root cause of that bug, this is the
real root cause.


src/perf_tool_backend.ml
```diff
@@ -156,13 +157,24 @@ module Recording = struct
       | None -> []
       | Some snapshot_size -> [ [%string "-m,%{Pow2_pages.num_pages snapshot_size#Int}"] ]
     in
+    let pid_opt = [ Pid.to_int pid |> Int.to_string ] in
+    let snapshot_opt =
+      if full_execution
```
Xyene, line 162: Hmm, is `full_execution` any different from `Magic_trace_or_the_application_terminates`? Could the code paths be merged? (Not asking you to do so in this PR)

Xyene (approved):
Good catch!

src/perf_tool_backend.ml
cgaebel: It is if the application runs longer than a snapshot buffer length.

Xyene: Ah, right.

PR #136, opened 2022-04-09
Type breakpoint address as `int64`

Xyene:
We don't currently allow putting breakpoints in kernel space, but
the types are nicer now.

Also:
- mark as `Pointer` in Fuchsia trace, so it displays as hex
- mark as `Int64.Hex.t` in `sexp_of` output


cgaebel (approved):
LGTM

PR #137, opened 2022-04-09
Drop dependency on `ocaml-probes`

Xyene:
We now do the `ptrace` dance ourselves.

### DIFF
--- a/src/trace.ml
+++ b/src/trace.ml
@@ -278,9 +278,16 @@ module Make_commands (Backend : Backend_intf.S) = struct
     let pid = Ptrace.fork_exec_stopped ~prog ~argv () in
     let%bind attachment = attach record_opts ~elf ~debug_print_perf_commands pid in
     Ptrace.resume pid;
-    Async_unix.Signal.handle Async_unix.Signal.terminating ~f:(fun signal ->
+    (* Forward ^C to the child, unless it has already exited. *)
+    let exited_ivar = Ivar.create () in
+    Async_unix.Signal.handle
+      ~stop:(Ivar.read exited_ivar)
+      Async_unix.Signal.terminating
+      ~f:(fun signal ->
         UnixLabels.kill ~pid:(Pid.to_int pid) ~signal:(Signal_unix.to_system_int signal));
     let%bind.Deferred (_ : Core_unix.Exit_or_signal.t) = Async_unix.Unix.waitpid pid in
+    (* This is still a little racey, but it's the best we can do without pidfds. *)
+    Ivar.fill exited_ivar ();
     let%bind () = detach attachment in
     return pid
   ;;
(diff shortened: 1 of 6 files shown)


src/trace.ml
```diff
@@ -273,14 +273,11 @@ module Make_commands (Backend : Backend_intf.S) = struct
...
-  let run_and_record record_opts ~elf ~debug_print_perf_commands ~command_with_args =
+  let run_and_record record_opts ~elf ~debug_print_perf_commands ~prog ~argv =
     let open Deferred.Or_error.Let_syntax in
-    let argv = Array.of_list command_with_args in
-    let pid = Probes_lib.Raw_ptrace.start ~argv |> Pid.of_int in
+    let pid = Ptrace.fork_exec_stopped ~prog ~argv () in
     let%bind attachment = attach record_opts ~elf ~debug_print_perf_commands pid in
-    Probes_lib.Raw_ptrace.detach (Pid.to_int pid);
-    Async_unix.Signal.handle Async_unix.Signal.terminating ~f:(fun signal ->
-        UnixLabels.kill ~pid:(Pid.to_int pid) ~signal:(Signal_unix.to_system_int signal));
```
Xyene, line 283: Deleting this code fixes this error on ^C of `run -serve`:

```
^C(monitor.ml.Error (Unix.Unix_error "No such process" kill "")
 ("Raised by primitive operation at Magic_trace_lib__Trace.Make_commands.run_and_record.(fun) in file \"src/trace.ml\", line 282, characters 8-88"
  "Called from Async_kernel__Job_queue.run_jobs in file \"src/job_queue.ml\", line 167, characters 6-47"
```

It's not necessary anymore because of the `prctl` that prevents the child from outliving magic-trace.

Xyene: Actually, hmm, it might still be required if you want to snapshot a trace on ^C... but I'm not sure if that functionality is useful for users using `run`?

cgaebel: The way I envisioned this is that ctrl+c kills the program being magic-traced, and that in turn makes perf take a snapshot if it hasn't already taken one.

cgaebel: I kind of think you should keep the signal handler but ignore errors from kill.

cgaebel (changes requested):
Just the signal handler thing. Other than that, LGTM.

src/trace.ml
Xyene: I think that makes sense. I've backed out the change for now.

I don't think we should unconditionally be killing, though. `-serve` may be running a long time (possibly as root, if `-trace-include-kernel`), in which case we might end up killing an arbitrary process. We should probably be resetting the signal handler, but it looks like `Async_unix` prevents doing that nicely with

```ocaml
val handle_default : [ `Do_not_use_with_async ] -> _
```

Xyene: Fixed now.

cgaebel (approved):
lgtm

PR #138, opened 2022-04-09
Detect when users don't have snapshot-at-exit

cgaebel:
Tristan mentioned that only relatively recent perfs support
snapshot-at-exit. This feature dynamically detects the user's perf
version and fails with a helpful error if they use an at-exit
trigger mode with an old version of perf.

Technically we say in the wiki that we *require* v5.4+, but I didn't
enforce that anywhere. I figure that if people really want to use
function-triggered snapshotting with an old version of perf, they should
be free to do so.

### DIFF
--- a/src/perf_tool_backend.ml
+++ b/src/perf_tool_backend.ml
@@ -66,6 +66,7 @@ module Recording = struct
   type t =
     { can_snapshot : bool
     ; pid : Pid.t
+    ; capabilities : Perf_capabilities.t
     }
 
   let perf_selector_of_trace_mode : Trace_mode.t -> string = function
@@ -97,6 +98,7 @@ module Recording = struct
   let attach_and_record
       { Record_opts.multi_thread; full_execution; snapshot_size }
       ~debug_print_perf_commands
+      ~(subcommand : Subcommand.t)
       ~(when_to_snapshot : When_to_snapshot.t)
       ~(trace_mode : Trace_mode.t)
       ~(timer_resolution : Timer_resolution.t)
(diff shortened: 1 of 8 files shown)


src/perf_tool_backend.ml
```diff
@@ -111,6 +111,17 @@ module Recording = struct
...
           "magic-trace must be run as root in order to trace the kernel"
     in
+    let%bind.Deferred.Or_error () =
+      match when_to_snapshot with
+      | Magic_trace_or_the_application_terminates ->
+        if Perf_capabilities.(do_intersect capabilities snapshot_on_exit)
+        then return (Ok ())
+        else
+          Deferred.Or_error.error_string
+            "magic-trace requires perf userspace tools > 5.4.\n\
```
Xyene, line 121: We should clarify that only this particular feature requires 5.4; we have graceful degradation elsewhere.

Xyene (approved):
Otherwise LGTM

src/perf_tool_backend.ml
Xyene: Actually, this only affects "application terminates" snapshotting, we should still allow "magic-trace exits' snapshotting (i.e. via ^C)

Xyene: Tangentially, I didn't get stderr from perf when I ran with `--snapshot=e` on a box that didn't support it

```
perf record -o /tmp/magic_trace.tmp.595b95/perf.data --event=intel_pt/cyc=1,cyc_thresh=1,mtc_period=0,noretcomp=1/u --timestamp --per-thread -t 48287 --snapshot=e
[Attached! Press Ctrl-C to stop recording]
^C[Got signal, detaching...]
[Finished recording!]
[Snapshot taken!]
(Unix.Exit (Exit_non_zero 255))
```

cgaebel: At the moment, this affects both applications-terminates snapshotting and magic-trace-exits snapshotting because we sigint instead of sigusr2 on ctrl+c.

I can imagine us going back to sigusr2 if perf is too old, but it's going to be pretty annoying to test changes against both code paths. I'd rather just say "you gotta use symbol snapshotting if you have an old perf" or "you can't use this tool at all if you have an old perf".

I wonder if there's a clever way we could bundle a new perf (with musl!?) with magic-trace...

cgaebel: BTW, there is no stderr from perf when you pass `--snapshot=e` when it doesn't support that. That's a perf bug, not a magic-trace bug.

Xyene (approved):
LGTM

PR #141, opened 2022-04-11
Serve trace files out of `/proc/self/fd`

Xyene:
Write to and serve from an indirect reference to `t.store_path`, through
our process' fd table.

This is a little grotesque, but avoids a race where the user runs
magic-trace twice with the same `-output`, such that the filename
changes under `-serve` -- without this hack, the earlier magic-trace
serving instance would start serving the new trace, which is unlikely to
be what the user expected.


cgaebel (approved):
LGTM

PR #143, opened 2022-04-11
Add `-min-trigger-time` to delay installing breakpoint

### LINKED ISSUE #56: common events trigger too fast
If I run a particular trace for an event I expect to be trigger very often (say expected 1000Hz) then as soon as magic trace starts recording, it stops. As a consequence I get a very limited history.

While my software trigger could be written to only probabalistically trigger I'd like it if magic trace had an option "refuse to trigger until we've been recording for TIME-SPAN".

Xyene:
This allows capturing traces with `-trigger` symbols that get called
with very high (possibly bursty) frequency, where we otherwise migt
trigger before the snapshot buffer has a chance to fill up.

Closes #56.

### DIFF
--- a/src/trace.ml
+++ b/src/trace.ml
@@ -142,6 +142,7 @@ module Make_commands (Backend : Backend_intf.S) = struct
     type t =
       { backend_opts : Backend.Record_opts.t
       ; multi_snapshot : bool
+      ; min_trigger_span : Time.Span.t
       ; when_to_snapshot : When_to_snapshot.t
       ; record_dir : string
       ; executable : string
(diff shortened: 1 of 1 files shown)


cgaebel (approved):
LGTM. Consider leaving something in the FAQ pointing people to this solution. You don't actually have to do it, I am literally asking you to consider it.

atulyagh03:
Hi @Xyene, do these changes still work? I needed this feature for a project, so I tried copying its changes to a local clone (magic-trace version 1.2.1) but got this error while building:

magic-trace$ dune build @default  @fmt --auto-promote
File "src/trace.ml", line 294, characters 27-38:
294 |       ; min_trigger_span : Time.Span.t
                                 ^^^^^^^^^^^
Error (alert deprecated): module Core.Time
[since 2021-11] Use [Time_float] or [Time_float_unix] instead
File "src/trace.ml", line 390, characters 22-60:
390 |         let%bind () = Async_unix.after opts.min_trigger_span in
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error: This expression has type unit Deferred.t
       but an expression was expected of type
         'a Async_kernel__Deferred_or_error.t = 'a Or_error.t Deferred.t
       Type unit is not compatible with type
         'a Or_error.t = ('a, Error.t) result 


I am not familiar enough with OCaml to understand how to resolve the 2nd error. It is fixable?

Thanks

PR #145, opened 2022-04-13
Add tests based on saved perf script traces

cgaebel:
This gives us the ability to copy and paste perf script output to write
regression tests.


Xyene (approved):
LGTM

PR #147, opened 2022-04-14
Explicitly model start trace / end trace commands

cgaebel:
Before this change, they were incorporated into the event kind. This
change splits them out into their own variant so we can think about
them explicitly.
    
It also has the nice side effect of printing a useful error message
when a user discovers a new kind/trace combination we haven't
previously considered possible.
    
Testing: The hello-world-with-kernel-tracing test has no diff.

Depends on #145

### DIFF
--- a/core/event.ml
+++ b/core/event.ml
@@ -1,29 +1,16 @@
 open! Core
 
-module End_kind = struct
-  type t =
-    | Call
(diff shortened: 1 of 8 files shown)


Xyene (approved):
LGTM

PR #148, opened 2022-04-14
Handle decode errors properly

cgaebel:
This work is mostly split across two commits: one that adds decode error parsing using our existing broken code and add tests, and one that fixes that broken code and demonstrates fixed tests.

It embeds a fairly large refactor of trace_writer.ml because it was impossible to figure out what was going on without it.

Also, stop printing [inferred start time] in function names -- demote that to an arg, instead. It's helpful for debugging magic-trace itself, so relegate it to printing if MAGIC_TRACE_DEBUG is set.

Testing: Loaded up a trace with 115 (!) decode errors and it actually looked pretty good.

Depends on #147


cgaebel:
I need to regenerate the hello-world-with-kernel-tracing trace. There's a bunch of bogus spans near the end but I think they're only there because that trace was generated without decode errors.

core/event.ml
```diff
@@ -28,14 +27,59 @@ module Location = struct
...
+    ; src : Location.t
+    ; dst : Location.t
+    }
+  [@@deriving sexp]
+end
+
+module Decode_error = struct
+  type t =
+    { thread : Thread.t
+          (* The time is only present sometimes. I haven't figured out when, exactly, but my
```
Xyene, line 55:
```suggestion
      (* The time is only present sometimes. I haven't figured out when, exactly, but my
```

src/perf_tool_backend.ml
```diff
@@ -255,126 +255,177 @@ module Perf_line = struct
...
+  let classify line =
+    if String.is_prefix line ~prefix:" instruction trace error"
+    then Trace_error
+    else Ok_perf_line
+  ;;
+
+  let parse_time ~time_hi ~time_lo =
+    let time_lo =
+      (* In practice, [time_lo] seems to always be 9 decimal places, but it seems good
+           to guard against other possibilities. *)
```
Xyene, line 285:
```suggestion
         to guard against other possibilities. *)
```

core/event.ml
cgaebel: I cannot, ocamlformat says so.

Xyene: :cry:

Xyene (approved):
LGTM, good work

Issue #93, opened 2022-03-28
Speed up decoding


Xyene:
Looks like the scanfs are the most expensive bit, and could be stuck into the regex.

![p](https://user-images.githubusercontent.com/1403503/160461731-942b1c8a-927f-4a6b-9666-86dd237df424.png)

/cc @billduff re: https://github.com/janestreet/magic-trace/pull/78#discussion_r835919755 :)

PR #150, opened 2022-04-15
Speed up hex string parsing

Xyene:
For a particular trace I had on hand, this brings down decode time from
7s to 6s.

Refs #93.

### DIFF
--- a/src/perf_tool_backend.ml
+++ b/src/perf_tool_backend.ml
@@ -255,6 +255,41 @@ module Perf_line = struct
     | Some offset -> offset
   ;;
 
+  let int64_of_hex_string str =
+    (* Bit hacks for fast parsing of hex strings.
+     *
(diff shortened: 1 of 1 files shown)


src/perf_tool_backend.ml
```diff
@@ -290,8 +290,19 @@ module Perf_line = struct
         in
         let time_hi = Int.of_string time_hi in
         let int64_of_hex_string str =
```
cgaebel, line 292: move to a top-level function and add a couple unit tests?

cgaebel (changes requested):

src/perf_tool_backend.ml
Xyene: Sure, done

cgaebel (approved):
LGTM. Branch free *and lookup table free*; nice!

PR #156, opened 2022-04-15
Finesse decode errors back in time by 1ns

### LINKED ISSUE #155: Decode errors with the same timestamp as the next event create broken stacks
Screen Shot 2022-04-15 at 6 01 46 PM

In this example, `asm_exc_page_fault` isn't actually the root of the stack, that should be `main`.

What's going on here is that there's a decode error at the point that it breaks. The entry into `asm_exc_page_fault` happens with the same timestamp as the decode error. By looking at a test case for this trace, `BEGIN asm_exc_page_fault` happens *before* `BEGIN main`, because `BEGIN main` is output at the very end of the trace (it's a `start_event`) and `asm_exc_page_fault` is not (because it's not a `start_event`).

We either need to treat events with the same start time as the last decode error as start events, or fiddle with timestamps so that nothing can happen at exactly the same time as a decode error.

cgaebel:
A practical, but not perfect, fix for #155: If events happen with
the exact same timestamp as a decode error, stacks break.
Implement a hack to prevent that:

If and only if it doesn't reorder events, pretend the decode
error happened 1ns sooner than it actually did. That makes it have
a distinct timestamp from any following events that have the same
timestamp.

This hack fails, and the user will see broken stacks, when the
previous event has the same timestamp as the overflow error.
I haven't seen that happen before and I imagine it's quite
rare -- it would imply Intel PT was in an overflow state for an
exceedingly small amount of time.

Fixes #155.

Testing: Ran demo.c, got 5 decode errors, and the broken stacks were
fixed. Also, page_fault.ml test replicates this error condition and
it changed in the way I expect.

### DIFF
--- a/src/trace_writer.ml
+++ b/src/trace_writer.ml
@@ -18,6 +18,7 @@ module Mapped_time : sig
   val is_base_time : t -> bool
   val add : t -> Time_ns.Span.t -> t
   val diff : t -> t -> Time_ns.Span.t
+  val sub_one_ns : t -> t
 end = struct
   module T = struct
     type t = Time_ns.Span.t [@@deriving sexp, compare]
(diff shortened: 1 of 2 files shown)


Xyene (approved):
Very smooth, +1

Issue #115, opened 2022-04-01
Don't crash on the v8 demo


cgaebel:
The v8 demo is in `demo/demo.js`

Right now, attempts to magic-trace it crash while trying to parse perf output:

```
   "3435768/3435768 1283684.240499276:   int                      556263730105 v8::base::OS::Abort+0x15 =>     556263608d40 v8::internal::Snapshot::DefaultSnapshotBlob+0x57e00"))
```

or even:

```
   "3435768/3435768 1283760.029165481:   int                      55626368b83f v8::internal::Snapshot::DefaultSnapshotBlob+0xda8ff =>     5562635f7e80 v8::internal::Snapshot::DefaultSnapshotBlob+0x46f40"))
```

At this point, I'd normally go add parsing for software interrupts, `int`. But I don't understand what's going on here. Looking at V8's source code, `OS::Abort` is supposed to abort the process. But the process didn't abort! Also also, DefaultSnapshotBlob is pretty trivial function that doesn't look like it should be generating software interrupts.

cgaebel:
It might be a V8-ism, but I'd like to point out that these are C++ functions, not JS.

Xyene:
Let's just treat it as a jump for now, and look into *why* later.

Xyene:
We weren't able to reproduce this again, so closing for now.

PR #158, opened 2022-04-16
Alternate fix for #115

cgaebel:
Instead of moving decode times back by 1ns, it's a little bit cleaner
(and more complete) to move following events forward by 1ns.

Testing: Tests updated how I expect, and also saw a clear trace with
decode error + same tick events following.


Xyene (approved):
LGTM

PR #160, opened 2022-04-16
Fix decoding failure when `-working-dir` is a relative path

Xyene:
This was broken because we passed magic-trace cwd-relative paths, but
also changed the working directory of `perf script`. One of these had to
give, so we no longer run `perf script` inside the working directory.

This matches what we do for `perf record`.


cgaebel (approved):
LGTM

PR #163, opened 2022-04-16
Make --version display git version information

quantum5:
This PR adds the shell script `bin/gen-versions.sh` which generates an OCaml C stub file containing a strong `generated_hg_version` function which returns the output of `git describe --always --dirty --abbrev=7 --tags`. `core_kernel` will [call this function](https://github.com/janestreet/core_kernel/blob/236a05affa13cfc1a40926f896b883ed024f6a69/version_util/src/version_util.ml#L9-L18) when displaying the version.

Example output: `v0.0.3-56-gfbf5462`

Note that `core_kernel` provides [its own version of this function](https://github.com/janestreet/core_kernel/blob/236a05affa13cfc1a40926f896b883ed024f6a69/version_util/src/version_util_fallback_stubs.c#L12-L15) as a weak symbol, which our version overrides.

We then tell `dune` to generate this stub file by running the shell script and linking the resulting `version_stubs.c` into the executable.

### DIFF
--- a/bin/dune
+++ b/bin/dune
@@ -2,5 +2,14 @@
  (names magic_trace_bin)
  (public_names magic-trace)
  (libraries core magic_trace_lib core_unix.command_unix)
+ (foreign_stubs
+  (language c)
+  (names version_stubs))
  (preprocess
   (pps ppx_jane)))
+
+(rule
+ (target version_stubs.c)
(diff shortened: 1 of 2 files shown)


cgaebel (approved):
LGTM, thanks!

PR #165, opened 2022-04-16
Don't unconditionally serve UI if envvar is set

Xyene:
Ideally we'd allow:

- `-serve`: serve on 8080
- `-serve X`: serve on X
- ``: don't serve

...but I don't think `Command.t` can be made to accept this, so I went
back to having a separate `-port` flag.

### DIFF
--- a/src/tracing_tool_output.ml
+++ b/src/tracing_tool_output.ml
@@ -15,17 +15,19 @@ module Serve = struct
     match Env_vars.perfetto_dir with
     | None -> Command.Param.return Disabled
     | Some perfetto_ui_base_directory ->
-      let%map_open.Command port =
+      let%map_open.Command serve =
+        flag "serve" no_arg ~doc:" Host the magic-trace UI locally."
+      and port =
         let default = 8080 in
         flag
-          "serve"
+          "serve-port"
           (optional_with_default default int)
           ~doc:
             [%string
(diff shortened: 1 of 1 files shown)


src/tracing_tool_output.ml
```diff
@@ -15,17 +15,19 @@ module Serve = struct
...
     | None -> Command.Param.return Disabled
     | Some perfetto_ui_base_directory ->
-      let%map_open.Command port =
+      let%map_open.Command serve =
+        flag "serve" no_arg ~doc:" Host the magic-trace UI locally."
+      and port =
         let default = 8080 in
         flag
-          "serve"
+          "port"
```
cgaebel, line 23: What do you think about calling this `-serve-port` so it sorts next to `serve` in the help?

cgaebel (approved):
One minor comment. LGTM regardless of what you decide to do about it (even if that's nothing).

src/tracing_tool_output.ml
Xyene: I initially tried that, because ideally the sort order would be:

`-serve`
`-serve-port`

but `Command.t` sorted them in reverse order. Right now they sort

`-port`
`-serve`

so having them shorter seems nice, but I guess if we add any other flag that sorts in-between it'd be less nice...

I just switched back to `-serve-port`.

PR #168, opened 2022-04-16
Add a basic bug-report issue template

cgaebel:
I propose we land this immediately (assuming you agree with the general idea of having one of these), then make tweaks by clicking the "edit" button on github.


Xyene (approved):
LGTM

PR #174, opened 2022-04-17
Don't warn when snapshotting during full execution

cgaebel:
This warning isn't very helpful, it's really no big deal if the snapshot
symbol is hit during a full-execution run. This also fixes a minor bug
where the message is printed at the end of every full-execution complete
run because a snapshot is taken at the very end.


Xyene (approved):
LGTM

PR #175, opened 2022-04-17
permit IP of 0 when parsing decode errors

cgaebel:
Bill found this when testing a program with the latest magic-trace

### DIFF
--- a/core/event.ml
+++ b/core/event.ml
@@ -56,7 +56,7 @@ module Decode_error = struct
        Skylake test machine has it while my Tiger Lake test machine doesn't. It could
        easily be a difference between different versions of perf... *)
     ; time : Time_ns_unix.Span.Option.t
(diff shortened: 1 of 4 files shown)


Xyene:
This sounds like a `printf` misfeature... `#` only modifies non-zero values for some reason...

The code that generates this is https://github.com/torvalds/linux/blob/a1901b464e7e3e28956ae7423db2847dbbfb5be8/tools/perf/util/auxtrace.c#L1642-L1643

In a separate repro:

```c
#include <stdio.h>
#include <inttypes.h>
#define PRI_lx64 "l" PRIx64

int main(void) {
	printf("%#"PRI_lx64"\n", 0);
	printf("%#"PRI_lx64"\n", 1);
	return 0;
}
```

Outputs:

```
0
0x1
```

Xyene (approved):
LGTM but this makes me sad.

PR #177, opened 2022-04-17
Always sigusr2 to take snapshot

cgaebel:
I thought SIGINT would work, but it doesn't. I think perf means that it
will take a snapshot when the program under test exits, not when perf
exits...


Xyene (approved):
LGTM

PR #178, opened 2022-04-17
Copy edit magic-trace output

cgaebel:
- Put spaces around square brackets, like perf does
- Do not put warnings in square brackets, again to copy perf
- Use more punctuation
- Prefer periods to exclaimation marks
- Be less specific with how long decoding can take


Xyene (approved):
LGTM

PR #179, opened 2022-04-17
Make the default timer resolution Normal

cgaebel:
...to match the help text. As an aside, it's pretty confusing for the
default to not be called Normal.


Xyene (approved):
Good catch, that's definitely a typo.

PR #185, opened 2022-04-21
Stop using owee demangling

cgaebel:
Perfetto (and perf, for that matter) can demangle symbols themselves,
they don't need us to do it. Further, this demangler was always a little
bit incomplete -- it wouldn't tell you the line numbers of closures.

Depends on janestreet/perfetto#21


Xyene (approved):
LGTM

PR #186, opened 2022-04-21
Fix debian package build

quantum5:
For some reason, dune is trying to install the vendored packages. Explicitly specifying that dune install should install magic-trace fixes the build.

Example fixed run: https://github.com/quantum5/magic-trace/runs/6104035999?check_suite_focus=true


cgaebel (approved):
LGTM, thanks!

PR #187, opened 2022-04-21
Track and report icache/instruction page hits per frame

Xyene:
h/t @michoecho for the inspiration from
https://github.com/michoecho/perf2perfetto/.

This is likely not entirely accurate, especially around trace
starts and other weird parts of the state machine. But I think it's good
enough as a first approximation.

The UI could also be improved, but I haven't thought hard about it or
decided upon anything yet. Clark suggests we should maybe display
`Byte_units` instead of raw cacheline numbers, and I'd immediately agree
if it weren't for blowing up the interned string table size. We could
patch Perfetto to recognize this particular argument, and format in the
UI? We might also want to display this information in a track of its
own, maybe.

Problems for another day.


cgaebel:
Can you add some tests? I think it'd be easy to print enough information to prove this works in the existing perf script tests. It may also be useful to add some unit tests to `Callstack.t` but less confident about that one.

cgaebel:
I've a half-baked idea for visualizing this: put bytes-of-instructions on the X axis instead of time-in-nanoseconds. It'd make it easy to see at a glance which functions contribute the most to icache pollution, subject to a ton of caveats.

PR #188, opened 2022-04-22
Improve checks for fzf and perf

cgaebel:
- fzf support no longer spews to stderr on failure
- we actually check that the user has perf installed (ubuntu doesn't
  do it by default!)


cgaebel:
I'm landing this without review because I'd like to actually get this released, and Tudor is in SF right now.

PR #198, opened 2022-04-26
Update README.md: Getting started

RobbertH:
Moving the gcc linker option `-ldl` later.
With `gcc -ldl demo.c -o demo`, I got: 
```
/usr/bin/ld: /tmp/ccg6cKgo.o: in function `main':
demo.c:(.text+0x30): undefined reference to `dlopen'
/usr/bin/ld: demo.c:(.text+0x40): undefined reference to `dlerror'
/usr/bin/ld: demo.c:(.text+0x6d): undefined reference to `dlerror'
/usr/bin/ld: demo.c:(.text+0x80): undefined reference to `dlsym'
/usr/bin/ld: demo.c:(.text+0x89): undefined reference to `dlerror'
/usr/bin/ld: demo.c:(.text+0x11d): undefined reference to `dlclose'
collect2: error: ld returned 1 exit status
```

While `gcc demo.c -ldl -o demo` compiles.
Here's a stackoverflow answer that explains why this is happening:
https://stackoverflow.com/questions/11893996/why-does-the-order-of-l-option-in-gcc-matter/11894098#11894098

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ You can point magic-trace at a function such that when your application calls it
 
 # Getting started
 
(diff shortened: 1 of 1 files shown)


cgaebel:
Thank you!

PR #203, opened 2022-05-12
Fix parsing of never-ending loop trace errors

Xyene:
Saw this while tracing through self-modifying kernel code.


cgaebel (approved):
LGTM, thanks!

PR #204, opened 2022-05-12
Proper OCaml exception tracking

Xyene:
This is accomplished by having an OCaml compiler patch that tracks where
it emits pushtrap, poptrap, and entertrap directives. We can then walk
them as we execute the program to figure out the state of the OCaml
exception stack.

The old stacktrace-based approach is maintained for binaries that are
compiled without this section.


src/trace_writer.ml
```diff
@@ -655,6 +636,109 @@ let check_current_symbol
     Callstack.push thread_info.callstack location.symbol
 ;;
 
+module Ocaml_hacks : sig
```
cgaebel, line 639: Could you leave a comment on this module describing your hacks? Maybe include a link to the relevant ocaml/ocaml PR that added the elf notes somewhere.

src/trace_writer.ml
```diff
@@ -655,6 +636,109 @@ let check_current_symbol
...
+                call
+                  t
+                  thread_info
+                  ~time
+                  ~location:
+                    { instruction_pointer = 0L
+                    ; symbol_offset = 0
+                    ; symbol = From_perf [%string "[push trap @ %{addr#Int64.Hex}]"]
+                    }));
+      last_known_instruction_pointer := Some dst.instruction_pointer
```
cgaebel, line 738: Why did you choose a ref for last_known_instruction_pointer instead of using a mutable field?

core/ocaml_exception_info.mli
```diff
@@ -0,0 +1,18 @@
+type t
```
cgaebel, line 1: This module could use a comment. I had some pretty basic questions when I first read it like "what is a pushtrap?"

cgaebel (changes requested):
I'm very excited for this change.

src/trace_writer.ml
Xyene: Sure, done.

src/trace_writer.ml
Xyene: We were already doing this for `frames_to_unwind` and I didn't want to change that code in this PR.

core/ocaml_exception_info.mli
Xyene: Done 👍

cgaebel (approved):
LGTM, thanks!

PR #208, opened 2022-05-17
Allow saving raw IPT data to .sexp format, bypassing .fxt generation

### LINKED ISSUE #54: Command line flag to trace or attach without converting the results to Fuchsia trace
Other tools can use magic_trace library to decode events and convert to different formats.

Xyene:
Closes #54.

### DIFF
--- a/src/trace.ml
+++ b/src/trace.ml
@@ -115,14 +115,14 @@ let write_trace_from_events
   close_result
 ;;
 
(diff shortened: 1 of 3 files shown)


cgaebel (approved):
LGTM. I wish there was a way to un-CPS write_and-maybe-view, but I don't feel strongly enough about that to do anything about it.

PR #209, opened 2022-05-18
Display DSO if symbol is unknown

### LINKED ISSUE #173: Print DSO if symbol is unknown
`perf script` can print symbols' DSOs. That seems useful for enriching `[unknown]` calls so people can figure out what package they need debug symbols for.

To be really keen, print a warning and give them a copy+pasteable command to install the necessary debug symbols.

Xyene:
Closes #173.


cgaebel (changes requested):
Could you please add a test with real dsos in it? Feel free to just regenerate one of the hello-world tests.

Also maybe worth trying a dso with spaces or close parens in the path, at least to demonstrate the edges of where things break down.

shoffmeister:
FWIW, I cherry-picked the one real applicable commit on top of the current master (`git fetch git@github.com:Xyene/magic-trace.git track-dso && git cherry-pick 2578e7157256f399dd7cb07a6aaf7ae23e2668fd`) and get

image

from doing an attach (as root) on a /usr/libexec/Xorg process (which is a Linux X11 server process).

From what I can tell, the module path / name resolution is correct (also the implicit vdso).

I'd be happy to see this in an official build.

cgaebel (approved):
LGTM, thanks!

PR #218, opened 2022-06-06
Added capability to attach to multiple PIDs for tracing

Lamoreauxaj:
This pull request is to add the ability to have a single trace attach to multiple process ids. The interface changes to now allow `-p` to take multiple PIDs as a comma separated list which leaves the interface backward compatible. This is essentially just fed into perf which already supports tracing multiple processes. This feature is only supported by attach and there is not a method for run to run and attach to multiple processes. 

None of the other flags can be specified separately per process, but the trigger point (`-trigger`) is specified on the PID at the head of the list given since this requires the ELF file.

The logic for loading `Perf_map` was changed to allow loading a single map per process ID. For run / decode, a single perf map is still loaded either by filename or by PID.

### DIFF
--- a/core/backend_intf.ml
+++ b/core/backend_intf.ml
@@ -22,7 +22,7 @@ module type S = sig
       -> trace_mode:Trace_mode.t
       -> timer_resolution:Timer_resolution.t
       -> record_dir:string
-      -> Pid.t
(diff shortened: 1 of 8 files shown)


Xyene (approved):
LGTM, thanks!

PR #219, opened 2022-06-08
Added ability to trace CPU frequencies with perf

Lamoreauxaj:
Perf already reports CPU frequency through CBR (core-to-bus ratio) events. In this PR, I process these events and create a Counter track in the trace file which will show up in Perfetto. It seems to work as expected and I am able to get the CPU to downclock by using AVX instructions and generally I can see the demo just oscillate close to it's max clock speed.

A few things to note:
- There are two power events that get outputted by perf when I enable them through `perf script`. This gives psb and cbr events. I ignore the psb events (they are a heartbeat like event), but do recognize them along with cbr which we are interested in. I'm not sure if there are examples where CBR events do not give us full information on the CPUs frequency transitions and perf does allow more information to be collected, so it's possible there can be more information of use to us.
- I changed the abstraction about events from Perf to allow different types of Events, i.e. `Power`.
- Perfetto was also changed slightly to ensure this doesn't fail an assert because Perfetto seems to say `lastValues[0] = 0` for the counter track, see [here](https://github.com/janestreet/perfetto/pull/25).
- It displays above the thread it corresponds to in Perfetto (and will show one for every thread). The naming is set to 'CPU:freq (MHz)' which looks a little awkward, but was the default behavior from Perfetto. You can also toggle Perfetto to show different scales, for example min - max instead of from 0. 
- In order to enable this, you pass `-trace-power` right now. It might be reasonable to enable it by default, but this is how I currently set it.

### DIFF
--- a/core/event.ml
+++ b/core/event.ml
@@ -38,14 +38,30 @@ module Location = struct
 end
 
 module Ok = struct
+  module Trace = struct
+    type t =
+      { thread : Thread.t
+      ; time : Time_ns.Span.t
+      ; trace_state_change : Trace_state_change.t option [@sexp.option]
+      ; kind : Kind.t option [@sexp.option]
+      ; src : Location.t
+      ; dst : Location.t
+      }
+    [@@deriving sexp]
+  end
+
+  module Power = struct
+    type t =
+      { thread : Thread.t
+      ; time : Time_ns.Span.t
+      ; freq : int
+      }
+    [@@deriving sexp]
+  end
+
   type t =
-    { thread : Thread.t
-    ; time : Time_ns.Span.t
-    ; trace_state_change : Trace_state_change.t option [@sexp.option]
-    ; kind : Kind.t option [@sexp.option]
-    ; src : Location.t
-    ; dst : Location.t
-    }
+    | Trace of Trace.t
+    | Power of Power.t
   [@@deriving sexp]
 end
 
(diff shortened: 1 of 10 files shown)


core/event.ml
```diff
@@ -38,14 +38,30 @@ module Location = struct
...
+  module Power = struct
+    type t =
+      { thread : Thread.t
+      ; time : Time_ns.Span.t
+      ; freq : int
+      }
+    [@@deriving sexp]
+  end
+
   type t =
```
Xyene, line 62: If this type changes, we need to update the sexp-output version header too.

src/perf_tool_backend.ml
```diff
@@ -376,8 +392,34 @@ module Perf_line = struct
...
+      let pid = Int.of_string pid in
+      let tid = Int.of_string tid in
+      let time = parse_time ~time_hi ~time_lo in
+      (match kind with
+      | "cbr" ->
+        (* cbr (core-to-bus ratio) are events which show frequency changes. *)
+        Some
+          (Power
+             { thread =
+                 { pid = (if pid = 0 then None else Some (Pid.of_int pid))
```
Xyene, line 407: We should extract this logic to a helper, since it's used in four places now.

src/perf_tool_backend.ml
```diff
@@ -376,8 +392,34 @@ module Perf_line = struct
...
+                 ; tid = (if tid = 0 then None else Some (Pid.of_int tid))
+                 }
+             ; time
+             ; freq = Int.of_string freq
+             })
+      | _ -> None)
```
Xyene, line 413: Explicitly list "psb offs" here, and raise if we get anything else.

src/perf_tool_backend.ml
```diff
@@ -500,22 +545,25 @@ module Perf_line = struct
     (module struct
       open Core
 
-      let check s = to_event s ~perf_map:None |> [%sexp_of: Event.t] |> print_s
+      let check s =
+        to_event s ~perf_map:None |> Option.value_exn |> [%sexp_of: Event.t] |> print_s
+      ;;
```
Xyene, line 549: We should add a couple of tests here showing `cbr`/`psb offs` parsing, so we don't accidentally regress it.

test/perf_script.ml
```diff
@@ -41,6 +41,8 @@ let run ?(debug = false) ?ocaml_exception_info ~trace_mode file =
       write_duration_begin ~args ~thread ~name ~time;
       printf "->          END   %s\n" name
     ;;
+
+    let write_counter ~args:_ ~thread:_ ~name:_ ~time:_ : unit = printf "writing counter"
```
Xyene, line 45: We can just have this be a no-op, no print needed.

Lamoreauxaj: Now a no-op.

src/perf_tool_backend.ml
Lamoreauxaj: Added this.

src/perf_tool_backend.ml
Lamoreauxaj: There is now `maybe_pid_of_string` which I used for this.

core/event.ml
Xyene: https://github.com/janestreet/magic-trace/blob/782794dd1fa2df4f7f1b5d28ff4451f880318167/src/trace.ml#L120

Lamoreauxaj:
@Xyene I removed the flag and the PR should be merge-able.

Xyene (approved):
LGTM, thanks!

PR #220, opened 2022-06-11
Add protocol to URLs embedded in help text

shoffmeister:
Having recognizable URLs inside the help text helps humans and terminal (auto-highlighting) tooling alike

Signed-off-by: Stefan Hoffmeister <stefan.hoffmeister@econos.de>


shoffmeister:
Before this patch, I was actually thinking that there were stray pixels in the help text.

image

PR #222, opened 2022-06-13
Filter trace to range

pbaumbacher:
Adds a new flag `-filter "(<start> <stop>)"` to magic trace. With this flag, the trace is trimmed to only include events that occur between `start` (inclusive) and `stop` (exclusive) (#81). If either symbol doesn't appear in the trace, no output is produced. In multi-snapshot mode, the trace only includes regions between consecutive `start` and `stop` calls. The `-filter` flag is compatible with the same modes as the `-trigger` flag.

When producing a `.sexp` output, only events that occur in the filtered region are emitted. When producing a fuchsia `.fxt` output, the full available callstack around the filtered region is reconstructed. Active stack frames at the beginning of the filtered region have the time `start` was called as their inferred start time.

### DIFF
--- a/core/trace_filter.ml
+++ b/core/trace_filter.ml
@@ -1,0 +1,37 @@
+open! Core
+
+module Unevaluated = struct
+  type t =
+    { start_symbol : Symbol_selection.t
+    ; stop_symbol : Symbol_selection.t
+    }
+  [@@deriving fields]
+
+  (* Accepts '(foo bar)', returns { start_symbol = User_selected "foo"; stop_symbol = User_selected = "bar" } *)
+  let arg_type =
+    Command.Param.Arg_type.create (fun s ->
+        let start_symbol, stop_symbol =
+          Sexp.of_string s
+          |> Tuple2.t_of_sexp String.t_of_sexp String.t_of_sexp
+          |> Tuple2.map ~f:Symbol_selection.of_command_string
+        in
+        { start_symbol; stop_symbol })
+  ;;
+end
+
+type t =
+  { start_symbol : string
+  ; stop_symbol : string
+  }
+[@@deriving sexp, fields]
+
+let param =
+  let open Command.Param in
+  flag
+    "-filter"
+    (optional Unevaluated.arg_type)
+    ~doc:
+      "_ [-filter \"(<START> <STOP>)\"] restricts the output of magic-trace to events \
+       ocurring between consecutive occurences of START and STOP. If either function is \
+       not called in the trace, no output will be produced."
+;;
--- a/core/trace_filter.mli
+++ b/core/trace_filter.mli
@@ -1,0 +1,19 @@
+open! Core
+
+(* A specification of a trace filter for magic-trace *)
+
+module Unevaluated : sig
+  type t =
+    { start_symbol : Symbol_selection.t
+    ; stop_symbol : Symbol_selection.t
+    }
+  [@@deriving fields]
+end
+
+type t =
+  { start_symbol : string
+  ; stop_symbol : string
+  }
+[@@deriving sexp, fields]
+
+val param : Unevaluated.t option Command.Param.t
--- a/src/for_range.ml
+++ b/src/for_range.ml
@@ -1,0 +1,103 @@
+open! Core
+open! Async
+open! Import
+
+module Symbol_hit = struct
+  module Kind = struct
+    type t =
+      | Start
+      | Stop
+    [@@deriving sexp_of]
+  end
+
+  type t =
+    { kind : Kind.t
+    ; symbol : Symbol.t
+    ; time : Time_ns.Span.t
+    }
+  [@@deriving sexp_of]
+end
+
+let range_hit_times ~decode_events ~range_symbols =
+  let open Deferred.Or_error.Let_syntax in
+  let%bind { Decode_result.events; _ } = decode_events () in
+  Deferred.List.map events ~f:(fun events ->
+      let { Trace_filter.start_symbol; stop_symbol } = range_symbols in
+      let is_start symbol = String.(Symbol.display_name symbol = start_symbol) in
+      let is_stop symbol = String.(Symbol.display_name symbol = stop_symbol) in
+      Pipe.filter_map events ~f:(function
+          | Error _ | Ok { data = Power _; _ } -> None
+          | Ok { data = Trace trace; time; _ } ->
+            (match trace.kind with
+            | Some Call ->
+              let symbol = trace.dst.symbol in
+              if is_start symbol
+              then Some { Symbol_hit.kind = Start; symbol; time }
+              else if is_stop symbol
+              then Some { Symbol_hit.kind = Stop; symbol; time }
+              else None
+            | _ -> None)
+          | Ok { data = Sample { callstack }; time; _ } ->
+            List.rev callstack
+            |> List.fold ~init:None ~f:(fun acc call ->
+                   match acc, call with
+                   | None, { symbol; _ } when is_start symbol ->
+                     Some { Symbol_hit.kind = Start; symbol; time }
+                   | None, { symbol; _ } when is_stop symbol ->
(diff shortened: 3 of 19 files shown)


core/trace_filter.ml
```diff
@@ -0,0 +1,25 @@
+open! Core
+
+module Range = struct
+  type t = string * string [@@deriving sexp]
```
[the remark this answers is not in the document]
Xyene: Actually, if we do this, we don't need the extra `Range` module, and could call `of_command_string` for both left/right of the tuple here.

core/trace_filter.mli
```diff
@@ -0,0 +1,14 @@
+open! Core
+
+(* A specification of a trace filter for magic trace *)
```
Xyene, line 3:
```suggestion
(* A specification of a trace filter for magic-trace *)
```

tiny nit

test/test.ml
```diff
@@ -675,6 +682,103 @@ let%expect_test "enqueing events at start" =
   return ()
 ;;
 
+let%expect_test "my test" =
```
Xyene, line 685: Nit :)

test/test.ml
```diff
@@ -675,6 +682,103 @@ let%expect_test "enqueing events at start" =
...
+      add Call 13 "stop_trigger";
+      add Return 14 "base_fn";
+      add Call 15 "post_fn";
+      add Return 16 "base_fn";
+      add Return 17 "outer";
+      events ())
+  in
+  let%bind () =
+    dump_using_file ~range_symbols:("start_trigger", "stop_trigger") events
+    (* dump_using_file ( events) *)
```
Xyene, line 711:
```suggestion
```

src/trace_writer.ml
```diff
@@ -787,17 +796,64 @@ let end_of_trace (T t) =
...
+let rewrite_all_callstacks t ~(thread_info : _ Thread_info.t) ~time =
+  let inactive_callstacks =
+    thread_info.inactive_callstacks |> Stack.to_list |> List.rev
+  in
+  List.iter inactive_callstacks ~f:(fun callstack ->
+      rewrite_callstack t ~callstack ~thread_info ~time);
+  rewrite_callstack t ~callstack:thread_info.callstack ~thread_info ~time
+;;
+
+let maybe_start_hot_section t ~should_write ~time =
```
Xyene, line 825: `maybe_start_trimmed_region`?

src/trace_writer.ml
```diff
@@ -787,17 +796,64 @@ let end_of_trace (T t) =
...
+  if (not t.in_hot_section) && should_write
+  then (
+    Hashtbl.iter t.thread_info ~f:(fun thread_info ->
+        flush t ~to_time:time thread_info;
+        Deque.clear thread_info.start_events);
+    t.in_hot_section <- true;
```
Xyene, line 831: `in_trimmed_range?

Xyene (commented):
Good work! Could you also please interactive rebase over current master, and squash commits into logical groupings? Seems like a few commits unrelated to your work snuck in.

src/trace_writer.ml
pbaumbacher: filtered_region?

core/trace_filter.mli
pbaumbacher: updated

test/test.ml
pbaumbacher: updated

test/test.ml
pbaumbacher: updated

core/trace_filter.ml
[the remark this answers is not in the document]
pbaumbacher: Turned it into a record and changed things around a bit so [Unevaluated] is the inner module. I think it reads a bit better, but let me know if I'm missing something obvious and can get rid of the extra module altogether.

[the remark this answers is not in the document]
Xyene: I guess what I'm unclear on is why not have:

```ocaml
module Symbol_range = struct
  type t =
    { start_symbol : Symbol_selection.t
    ; stop_symbol : Symbol_selection.t
    }

  (* Accepts '(foo bar)', returns { start_symbol = User_selected "foo"; stop_symbol = User_selected = "bar" } *)
  let arg_type = ...
end
```

and not have the `Option.map` call.

src/perf_tool_backend.ml
```diff
@@ -705,6 +705,43 @@ module Perf_line = struct
                 (symbol_offset 0x0))))))) |}]
       ;;
 
+      let%expect_test "unknown symbol in DSO" =
```
Xyene, line 708: Is this test new? Why is there a diff here?

core/trace_filter.ml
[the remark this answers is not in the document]
Xyene: (This lets you get rid of `String_pair` I think.)

src/trace.ml
```diff
@@ -45,6 +45,46 @@ let create_elf ~executable ~(when_to_snapshot : When_to_snapshot.t) =
   | Magic_trace_or_the_application_terminates, _ | _, Some _ -> return (Ok elf)
 ;;
 
+let evaluate_symbol_selection ~symbol_selection ~elf ~header =
+  let open Deferred.Or_error.Let_syntax in
+  match elf with
```
Xyene, line 50: Weakly think this should be written as

```ocaml
let%bind.Deferred.Or_error () =
   match elf with
   | ...
in
```

to decrease nesting.

core/trace_filter.mli
```diff
@@ -0,0 +1,19 @@
...
+
+module Unevaluated : sig
+  type t =
+    { start_symbol : Symbol_selection.t
+    ; stop_symbol : Symbol_selection.t
+    }
+  [@@deriving fields]
+end
+
+type t =
```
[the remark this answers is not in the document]
pbaumbacher: It's returned by `evaluate_trace_filter` in trace.ml and gets passed around until the filter is actually applied.

src/trace.ml
pbaumbacher: Made that change, I agree it looks better

core/trace_filter.ml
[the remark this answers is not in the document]
pbaumbacher: Yeah that makes sense, wasn't thinking through what I was doing with the `Option.map`

src/for_range.ml
```diff
@@ -0,0 +1,88 @@
...
+      ~f:(fun (hits, in_filtered_region) event ->
+        let hits, in_filtered_region =
+          match hits with
+          | [] -> hits, in_filtered_region
+          | hd :: tl ->
+            (match event with
+            | Error _ | Ok (Power _) -> hits, in_filtered_region
+            | Ok (Trace ok) ->
+              (match
+                 ( Time_ns_unix.Span.(ok.time = hd.time)
```
Xyene, line 77: How about something like

```ml
let hits, in_filtered_region =
  match hits with
  | [] -> hits, in_filtered_region
  | hd :: tl ->
    (match event with
    | Ok (Trace { kind = Some Call; time; dst; _ })
      when Time_ns_unix.Span.(time = hd.time) &&
           Symbol.equal dst.symbol hd.symbol -> tl, not in_filtered_region
    | _ -> hits, in_filtered_region)
in
```

?

Xyene (commented):
Otherwise LGTM. I think if you rebased/squashed we could be good to merge.

pbaumbacher:
> Otherwise LGTM. I think if you rebased/squashed we could be good to merge.

I had to update the range logic a bit to handle `Sample` events. The most straightforward way to do this was just to treat any stack containing a range symbol as a `Symbol_hit`, but a consequence is that in sampling mode the filtered region begins at the _end_ of the range start function, instead of at the beginning in tracing mode. 

Making the behavior identical would require a bit more work: the control flow would have to be fairly different depending on if we're receiving sampling events or perf events. I think using sampling with the filter isn't likely to be a super common use-case, but let me know what you think.

Issue #213, opened 2022-05-26
`-multi-snapshot` leads to staircase traces


Xyene:
I think we don't blow sufficient state away at the end of every snapshot, so we end up inferring incorrect timestamps.

<details><summary>Screenshot</summary>

![multi-snapshot](https://user-images.githubusercontent.com/1403503/170562871-1f924107-370f-409a-9e92-3f2ed2500097.png)

</details>

Xyene:
Closed in https://github.com/janestreet/magic-trace/pull/224.

PR #224, opened 2022-06-15
Allow running with sampling if Intel PT is unavailable

Lamoreauxaj:
The goal of this PR is to allow magic-trace to run within an environment which is not on a physical Intel machine. This adds the ability to (if Intel PT is unavailable) to still run magic-trace and have it use perf with sampling instead.

There is now a flag `-sampling` which can enable this and if Intel PT is not present, then this happens automatically. The interface is mostly the same, however now `-timer-resolution` can specify as `Sample { freq }` and `-snapshot-size` is ignored (with warning) if passed. The traces will default to 512K it seems (approximately ~30ms it seems with max sampling frequency).

Some things to note:
- I added another `Event.t` type which is `Sample` and refactored this type to share information with other existing events types. There are processed in `trace_writer.ml` in a similar way to before, but just ret / call based on the current state of the callstacks.
- Perf gives multiple options for reconstructing the callgraphs which are fp, dwarf, and lbr. LBR is only supported on Intel and we check for support and use it if possible. Otherwise it defaults to fp which will work if the binary and associated libraries were compiled with `-fno-omit-frame-pointer`. Dwarf tends to lead to massive perf.data files and slow decoding, so this was the choice I made.
- I took out `Perf_line` from `perf_tool_backend.ml` and there is now a `perf_decode.ml` to refactor things a little bit. There is also decoding support which works on the sampling output from perf.
- In order to take a snapshot, I use `--switch-output=signal` however this will have perf outputting multiple timestamped `perf.data` files. Thus decoding now can handle multiple files and return multiple pipes. Additionally this makes it easier to reset trace writing state between each snapshot which ends up fixing #213 as a result.


lpw25:
I'm not sure it should fallback automatically. I expect people will often fail to notice they are using the wrong machine if it does that.

Lamoreauxaj:
Yeah that's fair. It does technically warn the user at least though. However if you think it would be better to exit instead, I can change this behavior.

lpw25:
Maybe a warning is enough, I'm not sure. I'll defer to @Xyene and @cgaebel's judgement.

core/event.mli
```diff
@@ -35,30 +35,25 @@ module Location : sig
...
-      }
+      | Trace of
+          { trace_state_change : Trace_state_change.t option
+          ; kind : Kind.t option
+          ; src : Location.t
+          ; dst : Location.t
+          } (** Represents an event collected from Intel PT. *)
+      | Sample of { callstack : Location.t list }
+          (** Represents event collected through sampling. *)
+      | Power of { freq : int } (** Power event collected by Intel PT. *)
```
Xyene, line 48: Nit: sort these grouped by sample type, i.e. have `Power` next to `Trace`.

src/perf_capabilities.ml
```diff
@@ -41,10 +41,13 @@ module Version = struct
...
-    In_channel.read_all "/sys/bus/event_source/devices/intel_pt/caps/psb_cyc"
-  in
-  String.( = ) cyc_cap "1\n"
+  try
+    let cyc_cap =
+      In_channel.read_all "/sys/bus/event_source/devices/intel_pt/caps/psb_cyc"
+    in
+    String.( = ) cyc_cap "1\n"
+  with
+  | _ -> false
```
Xyene, line 50: Worth leaving a comment as to why. I'm assuming it's because this file doesn't exist on AMD? (In which case, I think we should be matching on the `ENOENT` error only.)

src/perf_decode.ml
```diff
@@ -0,0 +1,669 @@
...
+    try Re.Group.all (Re.exec ok_perf_line_re line), Ok_perf_line with
+    | _ ->
+      (try Re.Group.all (Re.exec ok_perf_power_line_re line), Ok_perf_power_line with
+      | _ -> Re.Group.all (Re.exec ok_perf_sample_line_re line), Ok_perf_sample_line))
+;;
+
+let parse_time ~time_hi ~time_lo =
+  let time_lo =
+    (* In practice, [time_lo] seems to always be 9 decimal places, but it seems good
+         to guard against other possibilities. *)
```
Xyene, line 93: Nit: this module got de-indented so I think all the comments in this file have wonky spacing now.

src/perf_tool_backend.ml
```diff
@@ -63,6 +63,38 @@ let perf_fork_exec ?env ~prog ~argv () =
...
+  In_channel.read_all "/proc/sys/kernel/perf_event_max_sample_rate"
+  |> String.rstrip (* Strip off newline *)
+  |> Int.of_string
+;;
+
+(* This checks if pdcm flag is present in lscpu. This is necessary for LBR to work.
+   Although I couldn't ascertain that it is also sufficient. *)
+let supports_lbr () =
+  let open Deferred.Or_error in
+  let%bind.Deferred lscpu_result = Process.run ~prog:"lscpu" ~args:[] () in
```
Xyene, line 76: We can't rely on `lscpu` always being present, it is an optional package that doesn't get installed alongside `perf`. However, the data you need here is also present in `/proc/cpuinfo` in almost the same format.

src/perf_tool_backend.ml
```diff
@@ -63,6 +63,38 @@ let perf_fork_exec ?env ~prog ~argv () =
...
+  match lscpu_result with
+  | Error _ -> error_string "Unable to execute lscpu."
+  | Ok result ->
+    let parse_error = error_string "Unable to determine LBR support from lscpu." in
+    let flag_re = Re.Perl.re {|^Flags:\s+(\S.*)$|} |> Re.compile in
+    let lines = String.split ~on:'\n' result in
+    let flags =
+      Core.List.find_map lines ~f:(fun line ->
```
Xyene, line 84: Is the `Core.` prefix necessary here?

src/perf_tool_backend.ml
```diff
@@ -63,6 +63,38 @@ let perf_fork_exec ?env ~prog ~argv () =
...
+
+(* This checks if pdcm flag is present in lscpu. This is necessary for LBR to work.
+   Although I couldn't ascertain that it is also sufficient. *)
+let supports_lbr () =
+  let open Deferred.Or_error in
+  let%bind.Deferred lscpu_result = Process.run ~prog:"lscpu" ~args:[] () in
+  match lscpu_result with
+  | Error _ -> error_string "Unable to execute lscpu."
+  | Ok result ->
+    let parse_error = error_string "Unable to determine LBR support from lscpu." in
```
Xyene, line 80: Nit: move this down to where it is raised.

src/perf_tool_backend.ml
```diff
@@ -75,10 +107,15 @@ module Recording = struct
...
-    | High -> "cyc=1,cyc_thresh=1,mtc_period=0,noretcomp=1"
+  let perf_intel_pt_config_of_timer_resolution
+      : Timer_resolution.t -> string Deferred.Or_error.t
+    = function
+    | Low -> Deferred.Or_error.return ""
+    | Normal -> Deferred.Or_error.return "cyc=1,cyc_thresh=1,mtc_period=0"
+    | High -> Deferred.Or_error.return "cyc=1,cyc_thresh=1,mtc_period=0,noretcomp=1"
+    | Sample _ ->
+      Deferred.Or_error.error_string
+        "Timer resolution passed can only be used with sampling."
```
Xyene, line 118:
```suggestion
        "[-timer-resolution Sample] can only be used in sampling mode. (Did you forget [-sampling]?)"
```

src/trace.ml
```diff
@@ -421,7 +450,15 @@ module Make_commands (Backend : Backend_intf.S) = struct
...
     and trace_mode = Trace_mode.param
     and timer_resolution = Timer_resolution.param
-    and backend_opts = Backend.Record_opts.param in
+    and backend_opts = Backend.Record_opts.param
+    and use_sampling =
+      flag
+        "-sampling"
+        no_arg
+        ~doc:
+          "Instead of Intel PT will use perf with sampling if passed. If Intel PT is not \
```
Xyene, line 459:
```suggestion
          "Use stacktrace sampling instead of Intel PT. If Intel PT is not \
```

src/trace.ml
```diff
@@ -421,7 +450,15 @@ module Make_commands (Backend : Backend_intf.S) = struct
...
+           available, magic trace will default to this anyway."
```
Xyene, line 460:
```suggestion
           available, magic-trace will default to this."
```

src/trace_writer.ml
```diff
@@ -821,121 +845,142 @@ let write_event (T t) event =
...
-      Ocaml_hacks.ret_track_exn_data t thread_info ~time;
-      check_current_symbol t thread_info ~time dst
-    | None, Some Start ->
-      (* Might get this under /u, /k, and /uk, but we need to handle them all
+  | Ok
+      { Event.Ok.thread = _ (* Already used this to look up thread info. *)
+      ; time = _ (* Already in scope. Also, this time hasn't been [map_time]'d. *)
+      ; data
+      } ->
+    (match data with
```
Xyene, line 853: Nit: instead of having an inner match, perhaps this code would look a bit better (less indented) if the outer block matched on

```ocaml
| Ok
      { Event.Ok.thread = _ (* Already used this to look up thread info. *)
      ; time = _ (* Already in scope. Also, this time hasn't been [map_time]'d. *)
      ; data = Power { freq }
      } ->
```

etc.

test/perf_script.ml
```diff
@@ -74,17 +74,19 @@ let run ?(debug = false) ?ocaml_exception_info ~trace_mode file =
       in
       let should_print_perf_line (event : Event.t) =
         match event with
-        | Ok (Trace event) ->
-          (* Most of a trace is just jumps within a single function. Those are basically
+        | Ok event ->
```
Xyene, line 77: Same comment here, we can reduce nesting by matching

```ocaml
| Ok (Power _) | Ok (Sample _) -> false
```

Xyene (commented):
Looking good.

I think it'd also be worth it to add a larger test under `tests/`. Maybe a C hello-world, or similar.

core/backend_intf.ml
```diff
@@ -19,9 +19,11 @@ module type S = sig
       -> debug_print_perf_commands:bool
       -> subcommand:Subcommand.t
       -> when_to_snapshot:When_to_snapshot.t
+      -> multi_snapshot:bool
       -> trace_mode:Trace_mode.t
       -> timer_resolution:Timer_resolution.t
       -> record_dir:string
+      -> intel_pt:bool (** Select whether to use Intel PT or sampling. *)
```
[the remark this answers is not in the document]
Xyene: Hmm, I don't have strong opinions, but I'll lean towards no until someone asks for it. I can't really think of a case where I'd rather want higher-overhead DWARF samples over lower-overhead LBR.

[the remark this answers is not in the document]
Lamoreauxaj: Do you have an idea about what to call this parameter? Something like trace_mode seems natural but that refers to the selection for userspace / kernel already.

core/event.mli
pbaumbacher: Could also nest by perf mode, along the lines of `Pt (Trace _) | Pt (Power _) | Sample _` since any given pipe will consist only of (`Trace` and `Power`) or `Sample`

Lamoreauxaj: That does seem like a reasonable option. One thought though, is what if we supported power events which aren't part of Intel PT as well and used the same `Power` event? (it seems like you can get the same information through other perf events). And another thought, although we could group them under `Pt`, they aren't really being processed or handled together which makes me lean towards leaving them as is since it doesn't seem useful to pattern match only on `Pt` events.

src/perf_decode.ml
Lamoreauxaj: Should be fixed now

src/perf_tool_backend.ml
Lamoreauxaj: I refactored to just check if all CPUs in /proc/cpuinfo contain the flag pdcm. It's also now located in `perf_capabilities.ml`.

src/trace_writer.ml
Lamoreauxaj: Updated

test/perf_script.ml
Lamoreauxaj: Updated

core/backend_intf.ml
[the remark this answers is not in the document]
Xyene: Maybe we could (in a separate, preceding commit) rename `trace_mode` to `trace_scope`? Alternatively, this could be named a `collection_mode`, because sampling isn't tracing.

core/backend_intf.ml
```diff
@@ -1,5 +1,5 @@
-(** Backends, which do the Processor Trace recording, present this unified
-    interface in order to be exposed as commands which can generate traces. *)
+(** Backends, which do the trace recording, present this unified interface in
+   order to be exposed as commands which can generate traces. *)
```
Xyene, line 2:
```suggestion
    order to be exposed as commands which can generate traces. *)
```

src/perf_tool_backend.ml
```diff
@@ -138,26 +166,79 @@ module Recording = struct
...
+      | Stacktrace_sampling ->
+        (match timer_resolution with
+        | Low -> return [ "-F"; "1000" ]
+        | Normal -> return [ "-F"; "10000" ]
+        | High -> return [ "-F"; Int.to_string (max_sampling_frequency ()) ]
+        | Sample { freq } -> return [ "-F"; Int.to_string freq ]
+        | Custom _ ->
+          error_string
+            "[-timer-resolution Custom] can only be used with Intel PT. (Are you running \
+             on a physical intel machine without [-sampling]?)")
```
Xyene, line 182:
```suggestion
             on a physical Intel machine without [-sampling]?)")
```

src/callgraph_mode.ml
```diff
@@ -0,0 +1,24 @@
...
+  | Last_branch_record
+  | Frame_pointers
+  | Dwarf
+[@@deriving sexp]
+
+let param =
+  let open Command.Param in
+  flag
+    "-callgraph-mode"
+    (optional (Command.Arg_type.create (fun str -> t_of_sexp (Sexp.of_string str))))
```
Xyene, line 13: We should also allow people to specify short-forms here too, i.e. "lbr", "fp", "dwarf", to be familiar to people used to `perf`. If the sexp parsing fails, we can fallback to a lookup table.

src/callgraph_mode.ml
```diff
@@ -0,0 +1,24 @@
...
+    ~doc:
+      "When perf is running with sampling collection mode, this sets how perf should \
```
Xyene, line 15: In general, try to avoid mentioning perf directly in the help text. Using perf is an "implementation detail", and e.g. the direct backend doesn't shell out to perf, but would have the same constraints (and ideally the help text would continue to apply).

core/collection_mode.ml
```diff
@@ -0,0 +1,28 @@
+open! Core
+
+type t =
+  | Intel_processor_trace
+  | Stacktrace_sampling
+
+let select_collection_mode use_sampling =
```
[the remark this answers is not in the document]
Lamoreauxaj: Fair, I didn't worry about this since it was a local unexposed helper function. But I still changed it for clarity.

src/callgraph_mode.ml
Lamoreauxaj: Replaced perf with magic-trace.

src/callgraph_mode.ml
```diff
@@ -0,0 +1,32 @@
...
+  | str -> t_of_sexp (Sexp.of_string str)
+;;
+
+let param =
+  let open Command.Param in
+  flag
+    "-callgraph-mode"
+    (optional (Command.Arg_type.create of_string))
+    ~doc:
+      "When magic-trace is running with sampling collection mode, this sets how it \
```
Xyene, line 22:
```suggestion
      " When magic-trace is running with sampling collection mode, this sets how it \
```

I think this is necessary to not have the help text look wonky.

test/perf_script.ml
```diff
@@ -74,22 +75,27 @@ let run ?(debug = false) ?ocaml_exception_info ~trace_mode file =
...
+          if should_print_perf_line event
           then (
-            let event = Perf_tool_backend.Perf_line.to_event line |> Option.value_exn in
-            let event = adjust_event_time event in
-            (* Most of a trace is just jumps within a single function. Those are basically
-           uninteresting to magic-trace, so skip them to keep tests a little cleaner. *)
-            if should_print_perf_line event then printf "%s\n" line;
-            Trace_writer.write_event trace_writer event));
+            match lines with
+            | [ line ] -> printf "%s\n" line
```
Xyene, line 96:
```suggestion
            | [ line ] -> print_endline line
```

src/trace_writer.ml
```diff
@@ -821,24 +846,47 @@ let write_event (T t) event =
...
+    let calls = List.drop callstack (Stack.length thread_info.callstack.stack) in
+    let (_ : Event.Location.t list) =
+      List.init how_many_call ~f:Fn.id
+      |> List.fold ~init:calls ~f:(fun calls _ ->
+             match calls with
+             | head :: tail ->
+               call t thread_info ~time ~location:head;
+               tail
+             | [] ->
+               raise_s [%message "BUG: Unexpected behavior while parsing callstack."])
```
Xyene, line 882: I think we should print some more information as part of this message, to make inevitable bug reports easier to track down.

Lamoreauxaj: This raise_s no longer exists and the loop is just an iter (ended up awkward because it wasn't looping the entire list when written).

Xyene (approved):
LGTM, thanks!

PR #227, opened 2022-06-23
Improved speed of hex parsing in more places

Lamoreauxaj:
This addresses issue #93. Although #150 implemented fast hex string parsing, when I was magic-tracing magic-trace, I noticed this in the trace:
image

This section of the trace entirely occurs because of calling `Int.Hex.of_string offset` within `parse_symbol_and_offset`. And this tends to take around 200-250 ns. Since this occurs about once per line in the `perf.data` file, the trace I ran this on executed this around 1.8 million times which should save a few hundred ms (out of total time of a few seconds).

Here are the traces after and before this change (respectively) zoomed in on 3 decodes (which shows 3 executions of `parse_symbol_and_offset`) of a perf line containing an entry of the callstack sampled. The former is just below 2 us and the latter is around 2.6 us. However it is clear that most of the remaining time is now spent on evaluating regex in the former.

image
image

In regards to implementation, I extracted these functions out to a `Util` module and used them when possible else where in magic-trace (although only the offset calculation is expensive because it is called a lot, but it seems reasonable to use them if possible). I used first class modules to abstract around `Int` and `Int64`.

### DIFF
--- a/core/perf_map.ml
+++ b/core/perf_map.ml
@@ -12,9 +12,6 @@ let perf_map_re =
   Re.Posix.re {|^([0-9a-fA-F]+) ([0-9a-fA-F]+) (.*)$|} |> Re.compile
 ;;
 
-let hex_to_int s = Int.Hex.of_string ("0x" ^ s)
-let hex_to_int64 s = Int64.Hex.of_string ("0x" ^ s)
-
 let parse_line line =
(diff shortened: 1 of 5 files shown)


Xyene (approved):
Nice catch! Holding off on merging until you've rebased, but this looks good to me.

PR #228, opened 2022-06-24
Added warnings to `-callgraph-mode` and added option for stitched LBR.

Lamoreauxaj:
This feature has two changes related to `callgraph_mode`.

* Based on user feedback from #224, I added warnings when the `-callgraph-mode` defaults to either `Dwarf` or `Last_branch_record` to notify the user of potential drawbacks from such selection. These warnings only show if they do not explicitly set the flag.
* Added a fourth option to run with `Last_branch_record { stitched = true }` which calls `perf report --call-graph lbr` and `perf script --stitch-lbr`. There are a few design decisions here though. First I decided to add it as a variant of `Callgraph_mode` for simplicity of interface for the user. However this has the awkward effect of this now being a parameter needed for both decoding and recording (which `Backend_intf` doesn't support). I would think this could perhaps be useful in other scenarios and so instead of pulling this out to `trace.ml` I decided to add a way for `decode_events` to have access to a `Recording.Data.t` record. Open to alternative ideas though.

I also added a wiki page to document the options for `-callgraph-mode`. This is not in the PR, but here is the link to the wiki repo: https://github.com/Lamoreauxaj/magic-trace.wiki.git. See [this page](https://github.com/Lamoreauxaj/magic-trace/wiki/Sampling-callgraph-modes). If this PR is merged, this wiki page would need to be merged too (I referenced it in the cli help documentation). And since I referenced it, the domain would need to redirect to the wiki page.

### DIFF
--- a/core/backend_intf.ml
+++ b/core/backend_intf.ml
@@ -12,6 +12,10 @@ module type S = sig
   end
 
   module Recording : sig
+    module Data : sig
+      type t [@@deriving sexp]
+    end
+
     type t
 
     val attach_and_record
@@ -25,7 +29,7 @@ module type S = sig
       -> record_dir:string
       -> collection_mode:Collection_mode.t
       -> Pid.t list
-      -> t Deferred.Or_error.t
+      -> (t * Data.t) Deferred.Or_error.t
 
     val maybe_take_snapshot : t -> source:[ `ctrl_c | `function_call ] -> unit
     val finish_recording : t -> unit Deferred.Or_error.t
@@ -40,6 +44,9 @@ module type S = sig
   val decode_events
     :  ?perf_maps:Perf_map.Table.t
     -> debug_print_perf_commands:bool
+    -> recording_data:Recording.Data.t option
+         (** This parameter is passed to allow [decode_events] to depend
+             on information or configuration from [attach_and_record]. *)
     -> record_dir:string
     -> collection_mode:Collection_mode.t
     -> Decode_opts.t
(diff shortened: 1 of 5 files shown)


src/callgraph_mode.ml
```diff
@@ -20,13 +24,13 @@ let param =
...
       " When magic-trace is running with sampling collection mode, this sets how it \
-       should reconstruct callstacks. The options are Last_branch_record (lbr), Dwarf \
-       (dwarf), and Frame_pointers (fp). Will default to Last_branch_record if supported \
-       and Dwarf otherwise."
+       should reconstruct callstacks. The options are lbr/lbr-no-stitch/dwarf/fp or a \
+       sexp. Will default to lbr is available and dwarf otherwise. For more info: \
+       https://magic-trace.org/w/c"
 ;;
 
 let to_perf_string = function
```
Xyene, line 32: Hmm, maybe this should be of the form "to perf args", and contain the stiched lbr bits too.

Lamoreauxaj: Perhaps, the only odd thing is args are needed for perf script and for perf record. Maybe two functions would be better then?

Lamoreauxaj: I committed how I would do this if you want to look

core/backend_intf.ml
```diff
@@ -40,6 +44,9 @@ module type S = sig
   val decode_events
     :  ?perf_maps:Perf_map.Table.t
     -> debug_print_perf_commands:bool
+    -> recording_data:Recording.Data.t
+         (** This parameter is passed to allow [decode_events] to depend
```
Xyene, line 48: Nit: indenting (ignore if ocamlformat is generating this alignment...)

PR #229, opened 2022-06-24
Added comment explaining callstack order of Sample event

Lamoreauxaj:
Peter pointed this out, but this was unclear without having to read the code what order the callstack list was in and so I felt a doc comment would be well-served here.


Xyene (approved):
LGTM, thanks!

Issue #32, opened 2022-03-18
Implement a `perf script --dlfilter` backend


Xyene:
Currently, we spend a lot of time `sscan`-ing textual output from `perf script`, and filter out branches within the same symbol in OCaml.

On recent (5.14+) `perf` installs, we could offload this to `perf script --dlfilter`, and have a short native object that does the filtering for us. Ref https://man7.org/linux/man-pages/man1/perf-dlfilter.1.html.

cgaebel:
When we do this, we should make sure to maintain support for pre-5.14 perf. Older perfs are extremely common, and most people aren't willing to install a recent perf userspace to use with their older kernel.

PR #231, opened 2022-06-29
Added support to run perf with dlfilter which filters out jumps within same function

Lamoreauxaj:
This addresses #32. Since 5.14, perf supports passing `--dlfilter file.so`. This PR creates a filter which skips over any Intel PT events which are branches within the same function. This filter is implemented in `perf_dlfilter.c` and built into `perf_dlfilter.so` by dune and using ocaml-crunch compiled into the ocaml executable for use.

Some things to note:
* There is a custom dune rule which builds the shared object. `ocaml-crunch` was added as a dependency and now to build, `gcc` is also required.
* magic-trace will detect support in `perf_capabilities.ml` for `--dlfilter` by checking the perf version. This ensures compatibility with older perf versions.
* `perf_dlfilter.h` is the include header from perf which is required. It's possible to use the system version, but placing it in magic-trace will ensure one can build even if that include is not present.

How much faster is magic-trace using the dlfilter? When running on a demo program, on a 19.194 MB perf.data file, running `perf script` goes from 1.980s with 1308709 lines to 0.534s with 358954 lines. On a 180.123 MB perf.data file, it goes from 110.279s with 75715521 lines to 53.746s with 20747507 lines. This does not include magic-trace's decoding time, but this should scale linearly with the number of lines. 

A future feature however could be to use this dlfilter to also pass information to ocaml without having to parse text using regex. However this feature alone is still a significant improvement for now.

### DIFF
--- a/src/perf_capabilities.ml
+++ b/src/perf_capabilities.ml
@@ -7,6 +7,7 @@ let kernel_tracing = bit 1
 let kcore = bit 2
 let snapshot_on_exit = bit 3
 let last_branch_record = bit 4
+let dlfilter = bit 5
 
 include Flags.Make (struct
   let allow_intersecting = false
@@ -18,6 +19,7 @@ include Flags.Make (struct
     ; kernel_tracing, "kernel_tracing"
     ; kcore, "kcore"
     ; last_branch_record, "last_branch_record"
+    ; dlfilter, "dlfilter"
     ]
   ;;
 end)
@@ -94,6 +96,9 @@ let supports_kcore = kernel_version_at_least ~major:5 ~minor:5
 (* Added in kernel commit ce7b0e4, which made it into 5.4. *)
 let supports_snapshot_on_exit = kernel_version_at_least ~major:5 ~minor:4
 
+(* Added in kernel commit 291961f, which made it into 5.14. *)
+let supports_dlfilter = kernel_version_at_least ~major:5 ~minor:14
+
 let detect_exn () =
   let%bind perf_version_proc = Process.create_exn ~prog:"perf" ~args:[ "--version" ] () in
   let%map version_string = Reader.contents (Process.stdout perf_version_proc) in
@@ -105,4 +110,5 @@ let detect_exn () =
   |> set_if (supports_kcore version) kcore
   |> set_if (supports_snapshot_on_exit version) snapshot_on_exit
   |> set_if (supports_last_branch_record ()) last_branch_record
+  |> set_if (supports_dlfilter version) dlfilter
 ;;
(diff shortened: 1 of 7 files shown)


src/perf_capabilities.ml
```diff
@@ -93,6 +95,7 @@ let supports_kcore = kernel_version_at_least ~major:5 ~minor:5
 
 (* Added in kernel commit ce7b0e4, which made it into 5.4. *)
 let supports_snapshot_on_exit = kernel_version_at_least ~major:5 ~minor:4
+let supports_dlfilter = kernel_version_at_least ~major:5 ~minor:14
```
Xyene, line 98: Could you add the commit this was added in here, as a comment?

src/perf_dlfilter.c
```diff
@@ -0,0 +1,20 @@
...
+int filter_event(void *data, const struct perf_dlfilter_sample *sample,
+                 void *ctx);
+int filter_event_early(void *data, const struct perf_dlfilter_sample *sample,
+                       void *ctx) {
+  const struct perf_dlfilter_al *resolved_ip =
+      perf_dlfilter_fns.resolve_ip(ctx);
+  const struct perf_dlfilter_al *resolved_addr =
+      perf_dlfilter_fns.resolve_addr(ctx);
+  if (!resolved_ip || !resolved_ip->sym || !resolved_addr ||
+      !resolved_addr->sym || strcmp(resolved_ip->sym, resolved_addr->sym) == 0)
```
Xyene, line 17:
```suggestion
      !resolved_addr->sym || strcmp(resolved_ip->sym, resolved_addr->sym) == 0) }{
```

nit

src/perf_tool_backend.ml
```diff
@@ -46,6 +46,19 @@ module Record_opts = struct
   ;;
 end
 
+let write_perf_dlfilter () =
+  let error = Deferred.Or_error.error_string "Unable to write [perf_dlfilter.so]." in
+  try
+    match Perf_dlfilter.read "perf_dlfilter.so" with
```
Xyene, line 52: Let's put this in the `-working-dir`.

src/perf_capabilities.ml
Lamoreauxaj: Added

src/perf_tool_backend.ml
Lamoreauxaj: Do we want to leave this around after magic-trace runs though? Do you think it should delete the shared object at the end of execution?

Xyene (approved):
LGTM, thanks!

PR #234, opened 2022-06-29
Add support for magic-trace to sample and show other events such as cache misses and branch misses

Lamoreauxaj:
The goal of this PR is to allow support to sample other events from perf, e.g. `branch-misses`, `cache-misses`. Although this is a nice feature, the goal is also to already have support for processing these events through sampling, so that when we have PEBS-via-PT available, we can also have these events in the processor trace stream.

Now the user can pass `-events` along with a comma separated list of either names like `branch-misses` or sexps which allow the user to control the event period. These will then show up on separate tracks along with the actual trace as arrows representing every sample. For example:
image

I also drafted a [new wiki page](https://github.com/Lamoreauxaj/magic-trace/wiki/Sampling-extra-events) and referenced in the help docs for `-events` which would need to be updated along with this PR.

This feature is split up into a few commits as follows:

### Updated perf decoding to process event types and periods and use event names for classification

`perf script` has support to pass the field `event` which actually outputs the type of the event. This actually helps improve classification by not just matching the perf line against regex and instead parsing the name of the event. And this also helps us recognize the new sampled events which we would not be able to differentiate otherwise. This commit just reworks the decoding to support parsing these event names and also the periods that can be outputted. Additionally all the tests are updated to support this.

### Updated `Collection_mode` to allow passing additional sampled events through to perf

This commit changes `Collection_mode` to also allow passing the `-events` flag. This includes constructing the perf event configs and passing them to perf here. There was also a bit of refactoring where some of the config parsing was moved into `collection_mode.ml` and out of `perf_tool_backend.ml`.

### Updated `perf_decode.ml` to handle decoding additional sampled events

This commit adds new event type to `event.ml` and changes `perf_decode.ml` to have regex for parsing these new events. Additionally tests were added of these new perf outputs.

### Updated `trace_writer.ml` to write new sampled events

`trace_writer.ml` now is able to handle writing these new events. It keeps track of the types of events and creates new 'thread' tracks which represent them. Each sample is represented as a zero-duration slice (arrow). Additionally since now events can appear before the actual snapshot took place, this also flushes the trace up to the start time of the snapshot.

### Added decoding/trace writing tests for sampling extra events

This commit solely adds new tests which include both branch-misses and cache-misses as well as the Intel PT stream or a stacktrace sampled stream.


core/collection_mode.ml
```diff
@@ -1,28 +1,117 @@
...
+
+  let of_string = function
+    | "branch-misses" -> { config = Period 50; name = Branch_misses }
+    | "cache-misses" -> { config = Period 50; name = Cache_misses }
+    | str -> t_of_sexp (Sexp.of_string str)
+  ;;
+
+  let arg_type = Command.Arg_type.create of_string
+
+  let to_perf_event_config { config; name } =
```
Xyene, line 43: Stuff in core should be agnostic to backend, so we shouldn't have `perf`-specific bits here. These functions should live in `src/` (and they shouldn't rely on `to_string` from `core` files matching up to what `perf` expects).

core/collection_mode.ml
```diff
@@ -1,28 +1,117 @@
...
+         (Command.Arg_type.comma_separated ~unique_values:true Event.arg_type))
+      ~doc:
+        "EVENTS Select additional events which can be sampled as a comma separated list. \
+         Valid options are [cache-misses] or [branch-misses]. For more info: \
+         https://magic-trace.org/w/e"
+  in
+  { primary; extra_events }
+;;
+
+let to_perf_record_arg
```
Xyene, line 90:
```suggestion
let to_perf_record_args
```

even though we return a 1-item list, it's still a list.

Xyene: But, this should also live somewhere in `src/`.

Xyene (commented):
I think a fair bit of code ends up needing to be moved here, so I'll hold off on reviewing the rest for now :)

Lamoreauxaj:
@Xyene I refactored the perf specific things out of core

core/collection_mode.ml
```diff
@@ -1,28 +1,95 @@
...
+    ; precision : Precision.t
+    }
+  [@@deriving of_sexp]
+
+  let of_string = function
+    | "branch-misses" ->
+      { when_to_sample = Period 50; name = Branch_misses; precision = Maximum_possible }
+    | "cache-misses" ->
+      { when_to_sample = Period 50; name = Cache_misses; precision = Maximum_possible }
+    | str -> t_of_sexp (Sexp.of_string str)
```
Lamoreauxaj, line 51: @Xyene a couple things to think about here. One, what do you think are good defaults for precision and when_to_sample? Two, do you think there should be an alternative to the sexp in order to control when_to_sample / precision?

core/collection_mode.ml
```diff
@@ -1,28 +1,96 @@
...
-let select_collection_mode ~use_sampling =
+let select_primary_collection_mode ~use_sampling =
   match use_sampling with
-  | true -> Stacktrace_sampling
+  | true -> Primary.Stacktrace_sampling
   | false ->
     (match Core_unix.access "/sys/bus/event_source/devices/intel_pt" [ `Exists ] with
     | Ok () -> Intel_processor_trace
     | Error _ ->
       Core.printf
```
Xyene, line 69: I think we want `eprintf` here.

core/collection_mode.ml
```diff
@@ -1,28 +1,96 @@
...
+    | str -> t_of_sexp (Sexp.of_string str)
+  ;;
+
+  let arg_type = Command.Arg_type.create of_string
+end
+
 type t =
-  | Intel_processor_trace
-  | Stacktrace_sampling
+  { primary : Primary.t
```
Xyene, line 58: Hmm, "Primary" is a weird name for this.

core/event.mli
```diff
@@ -44,9 +44,14 @@ module Ok : sig
...
           } (** Represents an event collected from Intel PT. *)
       | Power of { freq : int } (** Power event collected by Intel PT. *)
-      | Sample of
+      | Stacktrace_sample of
           { callstack : Location.t list (** Oldest to most recent calls in order. *) }
-          (** Represents event collected through sampling. *)
+          (** Represents cycles event collected through sampling. *)
+      | Event_sample of
+          { location : Location.t
+          ; period : int
```
Xyene, line 52: I don't think `period` is quite the right name in this context. `count` is a bit more accurate, though it is very much a function of the selected period.

src/perf_decode.ml
```diff
@@ -69,6 +68,49 @@ let parse_time ~time_hi ~time_lo =
...
+      in
+      let event =
+        match event_name with
+        | "branches" -> `Branches
+        | "cbr" -> `Cbr
+        | "psb" -> `Psb
+        | "cycles" -> `Cycles
+        | "branch-misses" -> `Branch_misses
+        | "cache-misses" -> `Cache_misses
+        | _ -> raise_s [%message "Unexpected event type when parsing perf output."]
```
Xyene, line 105: Worth printing some extra debug info here as part of this message.

src/perf_decode.ml
```diff
@@ -240,41 +256,96 @@ let ok_perf_line_to_event ?perf_maps matches line : Event.Ok.t =
...
+    line
+    lines
+    name
+    : Event.t
+  =
+  let (location : Event.Location.t) =
+    match lines with
+    | [] ->
+      (match Re.Group.all (Re.exec perf_extra_sampled_event_re line) with
+      | [| _str; _; instruction_pointer; symbol_and_offset |] ->
```
Xyene, line 297:
```suggestion
      | [| _; _; instruction_pointer; symbol_and_offset |] ->
```

src/perf_decode.ml
```diff
@@ -240,41 +256,96 @@ let ok_perf_line_to_event ?perf_maps matches line : Event.Ok.t =
...
+      | results ->
+        raise_s
+          [%message
+            "Regex of perf event did not match expected fields" (results : string array)])
+    | lines -> List.hd_exn lines |> parse_callstack_entry ?perf_maps thread
+  in
+  Ok
+    { thread
+    ; time
+    ; data = Event_sample { location; period = Option.value_exn period; name }
```
Xyene, line 308: If this is unconditionally going to raise on a `None` period, this should be reflected in the type signature (don't take an option).

src/trace_writer.ml
```diff
@@ -9,7 +9,7 @@ let is_kernel_address addr = Int64.(addr < 0L)
    compensate, the trace writer subtracts the time of the first event from all time spans, producing
    what we call a "mapped time". Only mapped times may be written to the trace file. *)
 module Mapped_time : sig
-  type t = private Time_ns.Span.t [@@deriving sexp_of, compare]
+  type t = Time_ns.Span.t [@@deriving sexp_of, compare]
```
Xyene, line 12: We want to keep this type private so we don't accidentally pass a regular `Time_ns.Span.t` to a function that requires `Mapped_time.t`. If you have a place where this is definitely necessary, you can use the `:>` operator.

src/trace_writer.ml
```diff
@@ -111,6 +111,8 @@ module Thread_info = struct
     ; start_events : (Mapped_time.t * Pending_event.t) Deque.t
           (* When the last event arrived. Used to give timestamps to events lacking them. *)
     ; mutable last_event_time : Mapped_time.t
+    ; trace_pid : int
```
Xyene, line 114:
```suggestion
    ; trace_pid : Pid.t
```
?

core/event.mli
Lamoreauxaj: I think perf does actually call this field period though. I can change it to count though as it feels more accurate.

src/trace_writer.ml
```diff
@@ -901,6 +905,34 @@ let write_event (T t) event =
...
+          allocate_thread t ~pid:thread_info.track_group_id ~name:track_name)
+    in
+    let args =
+      Tracing.Trace.Arg.(
+        List.concat
+          [ [ "timestamp", Int (Time_ns.Span.to_int_ns (time :> Time_ns.Span.t)) ]
+          ; [ "symbol", String (Symbol.display_name location.symbol) ]
+          ; [ "addr", Pointer location.instruction_pointer ]
+          ; [ "count", Int count ]
+          ; Option.map (Event.thread outer_event).pid ~f:(fun pid ->
```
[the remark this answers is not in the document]
Lamoreauxaj: Updated

src/perf_decode.ml
```diff
@@ -240,41 +254,92 @@ let ok_perf_line_to_event ?perf_maps matches line : Event.Ok.t =
...
-      | matches, Ok_perf_power_line ->
-        ok_perf_power_line_to_event matches |> Option.map ~f:(fun event -> Ok event))
+      let header = parse_event_header first_line in
+      (match header with
+      | Trace_error -> Some (Error (trace_error_to_event first_line))
+      | Event { thread; time; period; event; remaining_line } ->
+        (match event with
+        | `Branches ->
+          Some (parse_perf_branches_event ?perf_maps thread time remaining_line)
+          (* cbr (core-to-bus ratio) are events which show frequency changes. *)
```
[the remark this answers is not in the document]
Lamoreauxaj: Updated

src/perf_decode.ml
```diff
@@ -69,6 +68,47 @@ let parse_time ~time_hi ~time_lo =
...
+        | "branches" -> `Branches
+        | "cbr" -> `Cbr
+        | "psb" -> `Psb
+        | "cycles" -> `Cycles
+        | "branch-misses" -> `Branch_misses
+        | "cache-misses" -> `Cache_misses
+        | _ ->
+          raise_s
+            [%message
+              "Unexpected event type when parsing perf output." (event_name : string)]
```
Xyene, line 103:
```suggestion
              "Unexpected event type when parsing perf output" (event_name : string)]
```

Xyene:
LGTM, thanks!

PR #235, opened 2022-06-30
Allow extra spacing in cbr regex to not crash on some perf outputs

Lamoreauxaj:
I noticed when perf outputs frequencies which are three digits, then the output includes some extra spaces (in two places). I just added some extra logic to the regex to avoid this issue anymore. And I added the failure test case.


Xyene (approved):
LGTM, thanks!

PR #238, opened 2022-07-01
Updated `-trigger` help to specify triggers only apply to first PID passed

Lamoreauxaj:
Lee pointed out there is no help message which actually states triggers can only be set within the first PID passed. This PR just adds to the help doc.


Xyene (approved):
LGTM, thanks!

PR #239, opened 2022-07-01
Fixed regex bug in CBR event when flags includes space

Lamoreauxaj:
magic-trace crashes on the added test case due to missing a space in the regex. This PR just fixes this.


Xyene (approved):
LGTM, thanks!

PR #240, opened 2022-07-05
Disable dlfilter when printing trace to sexp

Lamoreauxaj:
This was requested to still allow tools to consume magic-trace output without dlfiltering. Thus when writing events to a sexp file, magic-trace will disable dlfiltering such that all events are outputted. If writing to a fuschia trace file, dlfilter will still be enabled if possible since the events filtered are not useful in the ui.

### DIFF
--- a/core/backend_intf.ml
+++ b/core/backend_intf.ml
@@ -43,6 +43,9 @@ module type S = sig
 
   val decode_events
     :  ?perf_maps:Perf_map.Table.t
+    -> ?filter_same_symbol_jumps:bool
+         (** Whether to filter unnecessary events which are jumps within the same
+             function. Default [true]. *)
     -> debug_print_perf_commands:bool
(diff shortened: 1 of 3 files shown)


core/backend_intf.ml
```diff
@@ -43,6 +43,7 @@ module type S = sig
 
   val decode_events
     :  ?perf_maps:Perf_map.Table.t
+    -> ?filter:bool (** Whether to filter unnecessary events. Default [true]. *)
```
Xyene, line 46:
```suggestion
    -> ?filter_same_symbol_jumps:bool (** Whether to filter unnecessary events. Default [true]. *)
```

Lamoreauxaj: I renamed all the uses of `filter` now.

Xyene (approved):
LGTM, thanks!

PR #241, opened 2022-07-07
Added ability to trigger on addresses and line numbers

Lamoreauxaj:
This feature adds triggers of the format `addr:0xfffff` or `line:file.ml:99:15`. All triggers can be prefixed with either `symbol:`, `addr:`, or `line:`, but if none is given, it will default to a symbol if possible and otherwise parse it as a line number or address.

There is no feature to allow our fuzzy finding to work with the new options, but it could be possible to implement fuzzy finding on files for example.

### DIFF
--- a/core/elf.ml
+++ b/core/elf.ml
@@ -136,6 +136,25 @@ let matching_functions t symbol_re =
   !res
 ;;
 
+let traverse_debug_line ~f t =
+  Option.iter t.debug ~f:(fun body ->
+      let cursor = Owee_buf.cursor body in
+      let pointers_to_other_sections =
+        Owee_elf.debug_line_pointers t.all_elf t.sections
+      in
+      let rec load_table_next () =
+        match Owee_debug_line.read_chunk cursor ~pointers_to_other_sections with
+        | None -> ()
+        | Some (header, chunk) ->
+          let process header (state : Owee_debug_line.state) () =
+            if not state.end_sequence then f header state
+          in
+          Owee_debug_line.fold_rows (header, chunk) process ();
+          load_table_next ()
+      in
+      load_table_next ())
+;;
+
 let find_symbol t name =
   let some_name = Some name in
   with_return (fun return ->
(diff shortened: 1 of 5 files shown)


core/elf.ml
```diff
@@ -148,6 +167,72 @@ let find_symbol t name =
...
+    traverse_debug_line
+      ~f:(fun header state ->
+        let filename = Owee_debug_line.get_filename header state in
+        let line = state.line in
+        let col = state.col in
+        match desired_filename, desired_line, filename with
+        | Some desired_filename, Some desired_line, Some filename ->
+          if String.(desired_filename = filename) && desired_line = line
+          then cols := (col, state.address) :: !cols
+        | None, _, _ | _, None, _ | _, _, None -> ())
```
Xyene, line 190:
```suggestion
        | _, _, _ -> ())
```

core/elf.ml
```diff
@@ -148,6 +167,72 @@ let find_symbol t name =
...
+    match cols with
+    | [] -> None
+    | (col, address) :: _ ->
+      (match desired_col with
+      | None ->
+        if List.length cols > 1
+        then
+          Core.eprint_s
+            [%message
+              "Multiple snapshot symbols on same line. Selecting one."
```
Xyene, line 204: Let's switch this to a regular `eprintf` because it's user-visible.

core/elf.ml
```diff
@@ -148,6 +167,72 @@ let find_symbol t name =
...
+  in
+  let find_addr_selection name =
+    Option.bind
+      (Option.try_with (fun () -> Int.Hex.of_string name))
+      ~f:(fun address -> Some (Selection.Address { address; name }))
+  in
+  let find_symbol_selection name =
+    Option.map (find_symbol t name) ~f:(fun symbol -> Selection.Symbol symbol)
+  in
+  if String.is_prefix name ~prefix:"symbol:"
```
Xyene, line 224: Let's extract this if/then branch into a short helper and use it here, that takes `~prefix` and `~f`, so that we can elide the hardcoding of the prefix lengths.

Xyene:
Closing since this ended up being merged as part of another PR.

PR #245, opened 2022-07-12
Added ability to fuzzy find on line number based triggers

Lamoreauxaj:
DRAFT: Rebased on #241.

This feature adds the ability to specify trigger modes like `-trigger line:?` and `-trigger symbol:?` which will cause fzf to only shows files or functions respectively. If `-trigger ?` is passed, it will show both. However if the user selects a file, there is another prompt for them to select the source location to break on.

### DIFF
--- a/core/dune
+++ b/core/dune
@@ -1,7 +1,7 @@
 (library
  (name magic_trace_core)
  (public_name magic-trace.magic_trace_core)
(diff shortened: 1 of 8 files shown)


Xyene (approved):
LGTM, thanks!

PR #247, opened 2022-07-14
destinations.ml: add a compressed_file_destination that uses zstandard

Lilydemi:
Trace files can be large and hard to share. Using zstd to compress trace
files allows users to decrease their file sizes, as demonstrated here:

```
$ du -sh trace.fxt trace.fxt.zst
472K    trace.fxt
56K     trace.fxt.zst
```

This is a preliminary attempt at adding a compressed file destination.
We have some open questions that we want to raise in the PR.

One thing we had to do is we had to make an `original_filename` labeled
argument. This is because, sometimes, magic-trace uses
`/proc/self/fd/$FD` paths as a way to prevent a race condition when
serving traces.

### DIFF
--- a/vendor/tracing/zero/destinations.ml
+++ b/vendor/tracing/zero/destinations.ml
@@ -1,9 +1,8 @@
 open! Core
 
-
-let direct_file_destination ?(buffer_size = 4096 * 16) ~filename () =
+let direct_file_destination ?(buffer_size = 64 * 1024) ~filename () =
   let buf = Iobuf.create ~len:buffer_size in
-  let file = Core_unix.openfile ~mode:[ O_CREAT; O_TRUNC; O_RDWR ] filename in
+  let file = Core_unix.openfile ~mode:[ O_CREAT; O_TRUNC; O_CLOEXEC; O_RDWR ] filename in
   let written = ref 0 in
   let flush () =
     Iobuf.rewind buf;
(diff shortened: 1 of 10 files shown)


vendor/tracing/zero/destinations.ml
```diff
@@ -32,7 +32,69 @@ let direct_file_destination ?(buffer_size = 4096 * 16) ~filename () =
...
+        ~len:com_size
+        (Iobuf.Expert.buf compressed_buf)
+    in
+    let compressed_length =
+      Zstandard.With_explicit_context.compress
+        compression_context
+        ~compression_level:com_level
+        ~input
+        ~output
+    in
```
Lilydemi, line 65: We (hlian and I) were uncertain how to reduce code duplication between `compressed_file_destination` and `file_destination`, so we wanted to open this PR up early. We think it might make sense to have a match statement in `let flush` and then _both_ uncompressed and compressed file destinations can use the same function. Something like

```ocaml
(* all the set up *)
let flush () =
  match is_compressed with
  | true -> ...
  | false -> ...
in
(* ... *)
(module Dest : Destination.S)
```

Does this make sense? If so, we can keep pushing code to this branch. Or, perhaps, is there a better way?

vendor/tracing/zero/destinations.ml
```diff
@@ -32,7 +32,69 @@ let direct_file_destination ?(buffer_size = 4096 * 16) ~filename () =
   (module Dest : Writer_intf.Destination)
 ;;
 
-let file_destination ~filename () = direct_file_destination ~filename ()
+let compressed_file_destination ?(buffer_size = 4096 * 16) ~filename () =
+  let buf = Iobuf.create ~len:buffer_size in
+  let com_size = buffer_size * 2 in
+  let com_level = 5 in
+  let compressed_buf = Iobuf.create ~len:com_size in
+  let file = Core_unix.openfile ~mode:[ O_CREAT; O_TRUNC; O_RDWR ] filename in
```
Xyene, line 40:
```suggestion
  let file = Core_unix.openfile ~mode:[ O_CREAT; O_TRUNC; O_CLOEXEC; O_RDWR ] filename in
```

In case this ends up being used with a trace processor shell-based `-serve`.

vendor/tracing/zero/destinations.ml
```diff
@@ -32,7 +32,69 @@ let direct_file_destination ?(buffer_size = 4096 * 16) ~filename () =
   (module Dest : Writer_intf.Destination)
 ;;
 
-let file_destination ~filename () = direct_file_destination ~filename ()
+let compressed_file_destination ?(buffer_size = 4096 * 16) ~filename () =
+  let buf = Iobuf.create ~len:buffer_size in
+  let com_size = buffer_size * 2 in
```
Xyene, line 37: Some of these variables use `com_` to mean compressed, while others use `compressed_`. We should use one for consistency; the latter sounds best for readability.

Xyene, line 37: Why does the compressed `Iobuf` size need to be 2&times; the size of of the regular buffer? Can we leave a comment here?

vendor/tracing/zero/destinations.ml
```diff
@@ -32,7 +32,69 @@ let direct_file_destination ?(buffer_size = 4096 * 16) ~filename () =
   (module Dest : Writer_intf.Destination)
 ;;
 
-let file_destination ~filename () = direct_file_destination ~filename ()
+let compressed_file_destination ?(buffer_size = 4096 * 16) ~filename () =
```
Xyene, line 35: Hmm, is there a semantic reason why this is written as 4 KiB * 16? Would `64 * 1024` or `65536` make more sense here?

vendor/tracing/zero/destinations.ml
```diff
@@ -32,7 +32,69 @@ let direct_file_destination ?(buffer_size = 4096 * 16) ~filename () =
...
+    Iobuf_unix.write compressed_buf file;
+    written := 0;
+    Iobuf.reset buf;
+    Iobuf.reset compressed_buf
+  in
+  let module Dest = struct
+    let next_buf ~ensure_capacity =
+      flush ();
+      if ensure_capacity > Iobuf.length buf
+      then failwith "Not enough buffer space in [direct_file_destination]";
```
Xyene, line 77: Nit: this error message references a different function name than the one we're currently in.

vendor/tracing/zero/destinations.ml
Lilydemi: Ok, leaving it as is!

vendor/tracing/zero/destinations.ml
Lilydemi: Left a comment!

vendor/tracing/zero/destinations.ml
Lilydemi: fixed this

hlian:
@Xyene ready for rereview if you have free time!

vendor/tracing/zero/destinations.ml
Xyene: Sorry, I think this review got accidentally marked as resolved.

Lilydemi: Just added the O_CLOEXEC in direct_file_destination as well. Does that resolve this review?

vendor/tracing/zero/destinations.ml
```diff
@@ -32,7 +31,72 @@ let direct_file_destination ?(buffer_size = 4096 * 16) ~filename () =
   (module Dest : Writer_intf.Destination)
 ;;
 
-let file_destination ~filename () = direct_file_destination ~filename ()
+let compressed_file_destination ?(buffer_size = 64 * 1024) ~filename () =
+  let buf = Iobuf.create ~len:buffer_size in
+  let compressed_size = buffer_size * 2 in
```
cgaebel, line 36: Why not `Zstandard.compression_output_size_bound`, instead of `compressed_size*2`? One example where `compressed_size*2` is not quite right is if the file is empty for some reason (the zstd'd file will not also be empty).

cgaebel: Also, very minor side note: comments go above the line you're commenting on, not below.

vendor/tracing/zero/destinations.ml
```diff
@@ -32,7 +31,72 @@ let direct_file_destination ?(buffer_size = 4096 * 16) ~filename () =
...
+  (* N.B. The size is 2x the buffer because, due to e.g. headers,
+     Zstandard will sometimes write slightly more data than the
+     original buffer had. *)
+  let compression_level = 5 in
+  let compressed_buf = Iobuf.create ~len:compressed_size in
+  let file = Core_unix.openfile ~mode:[ O_CREAT; O_TRUNC; O_CLOEXEC; O_RDWR ] filename in
+  let written = ref 0 in
+  let compression_context = Zstandard.Compression_context.create () in
+  let flush () =
```
cgaebel, line 45: Does this compress the full trace file at once? Would it be easy to change this to stream the file out, instead? I worry that this may cause users with relatively large trace files to OOM for no good reason. In particular, those users are the most likely to care about compressing their trace files in the first place.

Xyene:
Merged as part of https://github.com/janestreet/magic-trace/commit/7ac635e41438248c3a3348c039dae00854f38342. Thanks!

PR #248, opened 2022-07-18
Abena mega trace

salaast:
- Add use-processor-exe and open-url flags to run_command in trace.ml to automate large trace file processing and ui url opening


src/trace.ml
```diff
@@ -561,6 +561,32 @@ module Make_commands (Backend : Backend_intf.S) = struct
     { Decode_opts.output_config; decode_opts; print_events }
   ;;
 
+  let open_url_flags =
```
Xyene, line 564: I think we should allow configuration of this through envvars in `env_vars.ml`, rather than as commandline parameters. These choices tend to be sticky, so it's nicer to not have users have to specify them over and over.

I'd say a `MAGIC_TRACE_WEB_UI_URL` that can be any string, and which defaults to `https://magic-trace.org`, would be sufficient here.

src/trace.ml
```diff
@@ -561,6 +561,32 @@ module Make_commands (Backend : Backend_intf.S) = struct
...
+    |> map ~f:(fun user_input ->
+           Option.map user_input ~f:(fun user_choice ->
+               match user_choice with
+               | "local" -> "http://localhost:10000"
+               | "magic" -> "https://magic-trace.org"
+               | "android" -> "https://ui.perfetto.dev"
+               | _ -> raise_s [%message "unkown user input" (user_choice : string)]))
+  ;;
+
+  let use_processor_flag =
```
Xyene, line 582: I think we should take the path via an envvar, e.g. `MAGIC_TRACE_PROCESSOR_SHELL_PATH`; then we can have an optional `-use-trace-processor-shell` flag if that envvar is set. See https://github.com/janestreet/magic-trace/blob/872c648000ecff99c2035d239df4a53a8c5f91a0/src/tracing_tool_output.ml#L15 for inspiration.

We may also want to consider just using the trace processor shell automatically, past a certain filesize, but I haven't given that much thought.

src/trace.ml
```diff
@@ -589,31 +620,53 @@ module Make_commands (Backend : Backend_intf.S) = struct
...
+           match output with
+           | `Sexp _ -> failwith "unimplemented"
+           | `Fuchsia store_path -> store_path
+         in
+         let%bind.Deferred () =
+           match trace_processor_exe with
+           | None ->
+             print_endline "warning: must use local processor on large trace files";
+             Deferred.return ()
+           | Some processor_path -> Async_shell.run processor_path [ "-D"; output_file ]
```
Xyene, line 667: Can we use something like https://github.com/janestreet/magic-trace/blob/872c648000ecff99c2035d239df4a53a8c5f91a0/src/perf_tool_backend.ml#L70 here, to avoid the trace processor shell outliving magic-trace, in case magic-trace is <kbd>Ctrl</kbd>+<kbd>C</kbd>'d. May want to extract it into a helper module like `process.ml` or something.

src/trace.ml
```diff
@@ -589,31 +620,53 @@ module Make_commands (Backend : Backend_intf.S) = struct
...
-               ?range_symbols
-               ~elf
-               ~trace_scope:opts.trace_scope
-               ~debug_print_perf_commands
-               ~record_dir:opts.record_dir
-               ~collection_mode:opts.collection_mode
-               decode_opts))
+                 decode_opts)
+         in
+         let%bind.Deferred () =
```
Xyene, line 650: Hmm, I don't think we should do this. Oftentimes we'll run `magic-trace attach ... -serve` through SSH, so there'd be no browser to open and this'd fail.

src/trace.ml
salaast: Once I implement the Env_vars configuration, should I use the commandline flag to indicate whether the user wants to display the ui, or should it always be displayed?

src/trace.ml
salaast: Do you recommend catching an error here or something else? I was unable to find a way to detect from inside the function whether there is actually a browser available.

src/trace.ml
Xyene: I think for now, we should have a separate flag for if the trace processor shell should be used, if available. Maybe in the future we can make the selection automatic using some heuristic based on the size of the trace file (and whether we expect a browser to safely open it), but let's hold off on that bit for now.

src/trace.ml
Xyene: I think we just shouldn't be opening the user's browser. We already print a link to the terminal for where the user should go to view their trace, and many terminals just allow clicking into it directly.

src/env_vars.ml
```diff
@@ -22,3 +22,8 @@ let debug = Option.is_some (Unix.getenv "MAGIC_TRACE_DEBUG")
    [--dlfilter], this environment variable allows the user to forcibly disable
    filtering. *)
 let no_dlfilter = Option.is_some (Unix.getenv "MAGIC_TRACE_NO_DLFILTER")
+
+(* Points to a filesystem path with a trace processor executable. 
+   If provided, the flag can be indicated to automatically run the processor 
+   at the path once the trace file is created *)
+let processor_path = Unix.getenv "MAGIC_TRACE_PROCESSOR_SHELL_PATH"
```
Xyene, line 29:
```suggestion
let processor_path = Unix.getenv "MAGIC_TRACE_TRACE_PROCESSOR_SHELL_PATH"
```

src/env_vars.mli
```diff
@@ -4,3 +4,4 @@ val debug : bool
 val perf_is_privileged : bool
 val perfetto_dir : string option
 val no_dlfilter : bool
+val processor_path : string option
```
Xyene, line 7:
```suggestion
val trace_processor_shell_path : string option
```

to be explicit

src/trace.ml
```diff
@@ -8,8 +8,10 @@ let supports_command command =
   Lazy.from_fun (fun () ->
       match Core_unix.fork () with
       | `In_the_child ->
-        Core_unix.close Core_unix.stdout;
-        Core_unix.close Core_unix.stderr;
+        (* gracefully hide perf outputs *)
```
Xyene, line 11: We should leave a comment here about what versions of perf we observed breakages when close-ing and under what circumstances, because it was surprising.

src/trace.ml
```diff
@@ -561,6 +563,41 @@ module Make_commands (Backend : Backend_intf.S) = struct
     { Decode_opts.output_config; decode_opts; print_events }
   ;;
 
+  module Enable = struct
```
Xyene, line 566:
```suggestion
  module Enable = struct
```

Nit: we should give this module a more descriptive name.

src/trace.ml
```diff
@@ -561,6 +563,41 @@ module Make_commands (Backend : Backend_intf.S) = struct
...
+    type enabled = { value : string }
+
+    type t =
+      | Disabled
+      | Enabled of enabled
```
Xyene, line 571:
```suggestion
      | Enabled of { value : string }
```

Let's inline this type.

src/trace.ml
```diff
@@ -561,6 +563,41 @@ module Make_commands (Backend : Backend_intf.S) = struct
...
+            ~doc:[%string "use the trace processor set in environment variables"]
+        in
+        if processor then Enabled { value = processor_shell_path } else Disabled
+    ;;
+  end
+
+  (* Same as [Caml.exit] but does not run at_exit handlers *)
+  external sys_exit : int -> 'a = "caml_sys_exit"
+
+  let call_trace_processor ?env ~prog ~argv () =
```
Xyene, line 590: Let's either generalize this with the `perf` version, or leave a `CR-someday` to do this later.

src/trace.ml
Xyene: Let's also give `value` a better name, like `trace_processor_shell_path`.

src/trace.ml
```diff
@@ -573,11 +610,14 @@ module Make_commands (Backend : Backend_intf.S) = struct
          magic-trace run -multi-thread ./program -- arg1 arg2\n\n\
          # Run a process, tracing its entire execution (only practical for short-lived \
          processes)\n\
-         magic-trace run -full-execution ./program\n")
+         magic-trace run -full-execution ./program\n\
+         # Run a process that generates a large trace file, magic-trace run ./program \
+         -use-trace-processor-shell\n")
```
Xyene, line 615: We should leave this out of the documentation for now, or maybe make the `-use-trace-processor-shell` part of it conditional on the environment variable being present. Definitely don't want to give help text for a flag that doesn't exist.

src/trace.ml
salaast: I am unsure which part of this function is perf-specific.

Xyene: None of it is, so ideally this'd be the same function as https://github.com/janestreet/magic-trace/blob/d51d797dfd197367423a402b8fae1d56f46bc329/src/perf_tool_backend.ml#L70 so we don't have two copies.

You don't need to do this in this PR, but it would be good to leave a note for the future.

src/trace.ml
```diff
@@ -561,6 +564,42 @@ module Make_commands (Backend : Backend_intf.S) = struct
...
+  module Trace_processor_config = struct
+    type t =
+      | Disabled
+      | Enabled of { trace_processor_shell_path : string }
+
+    let param =
+      match Env_vars.trace_processor_shell_path with
+      | None -> Command.Param.return Disabled
+      | Some processor_shell_path ->
+        let%map_open.Command processor =
```
Xyene, line 576:
```suggestion
        let%map_open.Command use_trace_processor_shell =
```

Nit: descriptive variable names.

src/trace.ml
```diff
@@ -561,6 +564,42 @@ module Make_commands (Backend : Backend_intf.S) = struct
...
 
+  module Trace_processor_config = struct
+    type t =
+      | Disabled
+      | Enabled of { trace_processor_shell_path : string }
+
+    let param =
+      match Env_vars.trace_processor_shell_path with
+      | None -> Command.Param.return Disabled
+      | Some processor_shell_path ->
```
Xyene, line 575:
```suggestion
      | Some trace_processor_shell_path ->
```

Nit: avoid having multiple variable names for the semantically-same thing. It's best to shadow the previously-declared name in such cases.

src/trace.ml
```diff
@@ -561,6 +564,42 @@ module Make_commands (Backend : Backend_intf.S) = struct
...
+        let%map_open.Command processor =
+          flag
+            "use-trace-processor-shell"
+            no_arg
+            ~doc:[%string "use the trace processor set in environment variables"]
```
Xyene, line 580:
```suggestion
            ~doc:[%string "use trace processor shell to view trace"]
```

Xyene:
There seem to be some conflicts with the base branch, could you please try rebasing to resolve them?

Issue #212, opened 2022-05-25
Show call sources in call stacks


Xyene:
Sometimes it is useful to know where a function was called from. We currently don't display this information.

parthrohit22:
Hi, I'd like to work on this one. Planned approach: in src/trace_writer.ml, when writing a call's duration_begin event (write_pending_event'), resolve the caller's return address (top of the callstack before the push) through the existing Elf.Addr_table lookup already used for the callee's line/col/file, and attach it as additional args (e.g. call_site_file/call_site_line/call_site_col) so the call source is visible in the trace viewer alongside the called function.

Will follow up with a PR once I have something working, tested against the existing sample fixtures in test/. Let me know if this conflicts with other in-flight work or if there's a preferred approach.

parthrohit22:
Opened #390 with an implementation of this - resolves the caller's return address through the existing `Elf.Addr_table` debug-info lookup (same mechanism the callee's location already uses) and attaches it as `call_site_*` args on the `duration_begin` event. Covered by a new expect-test in `test/test.ml`.

I wasn't able to build/run the full suite locally (this repo's C stubs are Linux-only and I'm on macOS), so flagging that in case CI turns up anything platform-specific I couldn't catch.

PR #249, opened 2022-07-19
Show call source in UI

lamoreauxaj:
For #212, this PR shows the source of each call in the args shown given we already have that information. Overall pretty simple, but I modified the calls to store where they were called from or whether it was unknown (thus also removing `from_untraced`). It shows the information as `source_` followed by each of symbol, offset, and address as different args.


cgaebel:
Hi Aaron. This looks great, but would you please fix the conflict before I merge it?

PR #250, opened 2022-07-20
Demangle OCaml symbols

ClyveGassant:
This addresses #82 by implementing the demangling parser from perf. I also included expect tests.

PR #251, opened 2022-07-25
.github/workflows/build.yml: bump upx from 3.96-2 to 3.96-3

hlian:
Should unbreak #250.


hlian:
It looks like 3.96-3 isn't the right way to do this -- it'll require upgrading the base Ubuntu image -- lemme see if i can troubleshoot my way out of this on my own repo first. I'll open a new PR soon

PR #252, opened 2022-07-26
.github/workflows/build.yml: download upx 3.96-2 from ftp.debian.org

hlian:
The old URL disappeared, but this new location appears to work (when I tested it on my personal GitHub repo's CI). I'm not sure how relevant the old comment about musl is -- I didn't investigate deeply. To cover all the bases, I tweaked the comment slightly to just indicate that we're not using the default Ubuntu UPX. I'm guessing the Debian package is just built differently from the Ubuntu equivalent, but, again, didn't investigate too deeply.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,7 +21,7 @@ jobs:
       - uses: actions/cache@v2
         with:
           path: ~/.opam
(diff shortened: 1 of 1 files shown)


Xyene (approved):
LGTM, thanks!

PR #253, opened 2022-08-03
Add separate events output in sexp or binio format

Istlemin:
This feature adds the argument -events-output=FILENAME and -callstacks.

Magic trace will dump all events to the events-output file,
either in sexp or binio format depending on the suffix.

If the flag -callstacks is set, for every event the current callstack will
also be dumped, using a compression algorithm to reduce size.


cgaebel:
Would you please rebase this onto the latest master before I take a look? It looks like this PR is based on code from March.

src/callstack_compression.mli
```diff
@@ -0,0 +1,34 @@
...
+  }
+[@@deriving sexp, bin_io]
+
+(* Compress a callstack represented as a list of symbol, and update the compression
+    state inplace.
+
+    When compressing a sequence of many callstacks, this function should be called
+    in order on all callstacks with the same compression state.
+*)
+val compress_callstack : state:t -> Symbol.t list -> compression_event
```
cgaebel, line 26: House style is to not label arguments of type `t`, unless there's a particularly good reason to.

test/callstack_compression_test.ml
```diff
@@ -0,0 +1,82 @@
+open Core
+open Magic_trace_lib
+open Import
+
+let compressed_test_string =
+  "(\n\
```
cgaebel, line 6: If you use raw strings (e.g. `{|raw string|}`), you won't need any backslashes in this test output.

src/tracing_tool_output.ml
```diff
@@ -113,33 +113,46 @@ module Serve = struct
...
-      ~doc:[%string "FILE Where to output the trace. (default: '%{default}')"]
+      ~doc:
+        [%string
+          "FILE Fuchsia file to output the fuchsia trace to. (default: '%{default}')"]
+  and events_path =
+    flag
+      "events-output"
+      (optional string)
+      ~doc:[%string "FILE Sexp or binio file to output the trace events to."]
+  and output_callstack =
```
cgaebel, line 149: When would a person want to enable this flag? When would they want to disable it? If the compiler feature you're working on always wants this enabled, just enable it by default and delete the flag.

src/tracing_tool_output.ml
```diff
@@ -113,33 +113,46 @@ module Serve = struct
...
     let default = "trace.fxt" in
     flag
       "output"
       (optional_with_default default string)
       ~aliases:[ "o" ]
-      ~doc:[%string "FILE Where to output the trace. (default: '%{default}')"]
+      ~doc:
+        [%string
+          "FILE Fuchsia file to output the fuchsia trace to. (default: '%{default}')"]
+  and events_path =
```
cgaebel, line 144: Instead of adding a new argument here, what do you think about making it a function of the file extension given to `store_path`? e.g. if the file ends with `.sexp` or `.binio`, use your logic; else, do what it currently does.

cgaebel (changes requested):
I'm still reviewing this. Just some initial comments for now while I take a break.

src/tracing_tool_output.ml
Istlemin: We could do this if we think it's important. The way the code is structured right now makes it a bit difficult as the code that reconstructs the call stacks is quite tightly integrated with the code that writes the fucshia file, and these would then need to be separated.

cgaebel: In test, there's a way to replace the fuchsia file writing with something that logs each entry. Can you use that same interface to replace it with a "null writer" that just drops the fuchsia file entirely?

src/tracing_tool_output.ml
```diff
@@ -153,34 +154,60 @@ let maybe_stash_old_trace ~filename =
...
-  | `Sexp store_path -> Writer.with_file_atomic store_path ~f:f_sexp
-  | `Fuchsia store_path ->
-    let fd = Core_unix.openfile store_path ~mode:[ O_RDWR; O_CREAT; O_CLOEXEC ] in
-    (* Write to and serve from an indirect reference to [store_path], through our process'
-     fd table. This is a little grotesque, but avoids a race where the user runs
-     magic-trace twice with the same [-output], such that the filename changes under
-     [-serve] -- without this hack, the earlier magic-trace serving instance would start
-     serving the new trace, which is unlikely to be what the user expected. *)
+  let { serve; output_path } = t in
+  match String.is_suffix ~suffix:".fxt" output_path with
```
gretay-js, line 170: Is there a naming convention for Fuchsia format files and for Binio files? I've seen both `.ftf` and `.fxt` used for Fuchsia. This PR requires it to be `.fxt` and gives an error if the output file name extension is not one `.sexp .fxt .binio`, whereas before there was no such restriction. Also, should it be ".bin" instead of ".binio"?

cgaebel: `.fxt` is the only file extension for fuchsia trace files. `.ftf` was present in pre-release magic-trace because we didn't know what the right file extension was.

I prefer `.binio` to `.bin`, but I don't know if there's a standard naming convention. I prefer `.binio` because people sometimes use `.bin` to refer to other things, whereas I've never seen `.binio` used for anything else.

src/trace.ml
```diff
@@ -108,16 +104,58 @@ let write_trace_from_events
...
     else events
   in
+  let trace =
+    let%map.Option writer = writer in
+    let base_time =
+      Time_ns.add (Boot_time.time_ns_of_boot_in_perf_time ()) earliest_time
+    in
+    Tracing.Trace.Expert.create ~base_time:(Some base_time) writer
+  in
+  let module Null_writer = struct
```
cgaebel, line 114: Please define this module top-level unless there's a reason for it to close over anything else. Also, annotate a signature to ensure that errors appear on the module when someone changes the sig instead of where it's used.

src/tracing_tool_output.mli
```diff
@@ -14,6 +24,9 @@ val param : t Command.Param.t
 val write_and_maybe_view
   :  ?num_temp_strs:int
   -> t
-  -> f_sexp:(Writer.t -> 'a Deferred.Or_error.t)
-  -> f_fuchsia:(Tracing_zero.Writer.t -> 'a Deferred.Or_error.t)
+  -> f:
+       (?events_writer:events_writer
```
cgaebel, line 28: I think this parameter should be an `option`, not "optional". You have to pass a closure that takes this optional argument either way, might as well make the type reflect that fact.

And I think it's right in the sense that `f` really should be thinking about how to deal with no events writer explicitly.

src/trace_writer.ml
```diff
@@ -875,10 +883,34 @@ let maybe_stop_filtered_region t ~should_write =
     t.in_filtered_region <- false)
 ;;
 
+let write_event_and_callstack events_writer event callstack =
+  let compression_event =
+    Callstack_compression.compress_callstack
+      Tracing_tool_output.(events_writer.callstack_compression_state)
```
cgaebel, line 889: If you annotate the argument `events_writer` with a type, you can replace this line with `events_writer.callstack_compression_state`, which I find a little cleaner. You can also remove the type annotation when matching on it.

src/trace_writer.ml
```diff
@@ -875,10 +883,34 @@ let maybe_stop_filtered_region t ~should_write =
...
+      |> Stack.to_list
+      |> List.map ~f:(fun Event.Location.{ symbol; _ } -> symbol))
+  in
+  let event_and_callstack =
+    Event_and_callstack.{ event; callstack = compression_event }
+  in
+  match events_writer with
+  | Tracing_tool_output.{ format = Sexp; writer; _ } ->
+    Async.Writer.write_sexp
+      ~terminate_with:Async.Writer.Terminate_with.Newline
```
cgaebel, line 900: Delete `Async.Writer.Terminate_with.`. The compiler can infer that.

src/trace_writer.ml
```diff
@@ -875,10 +883,34 @@ let maybe_stop_filtered_region t ~should_write =
...
+      writer
+      [%sexp (event_and_callstack : Event_and_callstack.t)]
+  | Tracing_tool_output.{ format = Binio; writer; _ } ->
```
cgaebel, line 903: Match on `events_writer.format`, not `events_writer`

src/trace_writer.ml
```diff
@@ -875,10 +883,34 @@ let maybe_stop_filtered_region t ~should_write =
...
+      writer
+      Event_and_callstack.bin_writer_t
+      event_and_callstack
+;;
+
 (* Write perf_events into a file as a Fuschia trace (stack events). Events should be
    collected with --itrace=be or cre, and -F pid,tid,time,flags,addr,sym,symoff as per
    the constants defined above. *)
-let write_event (T t) event =
+let write_event (T t) ?events_writer event =
```
cgaebel, line 913: The diff to this function is pretty hard to read. Can you explain in words what you changed here in a reply to this comment?

cgaebel (changes requested):

src/trace_writer.ml
Istlemin: Huh, I'm really surprised GIT shows the diff so poorly, I guess its because of the changed indentation. All that's changed here is that `event` is being matched on in two steps instead of one. Now it is first matched with `Ok event_value` or `Error ...`, then `write_event_and_callstack` is run, and then `event_value` is being matched further.

src/tracing_tool_output.ml
Istlemin: So do we think it's fine the way it is now or should I change something? For example, assume fuchsia by default and only 
change to binio or sexp if extension matches

cgaebel: I like "assume fuchsia by default"

src/trace_writer.ml
cgaebel: That makes sense, thanks.

cgaebel:
LGTM. Would you please squash your changes into a single commit before I merge this?

cgaebel (approved):

cgaebel:
Thank you!

PR #254, opened 2022-08-05
Perf capability check

salaast:
Use capability parser to parse the output of getcap command on the perf file path; determine permission to trace kernel based on the linux version and the existence of cap_sys_admin/cap_perfmon in effective capabilities

### DIFF
--- a/src/capability.ml
+++ b/src/capability.ml
@@ -1,0 +1,222 @@
+(** Stand in for C capability library; suggested to use that library once it is open source *)
+
+open Core
+
+module Capability = struct
+  type t =
+    { effective : bool
+    ; permitted : bool
+    ; inherited : bool
+    }
+  [@@deriving fields]
+
+  let create = Fields.create
+
+  (* let to_string t =
+    "eff="
+    ^ Bool.to_string t.effective
+    ^ " per="
+    ^ Bool.to_string t.permitted
+    ^ " inh="
+    ^ Bool.to_string t.inherited
+  ;; *)
+
+  let empty = create ~effective:false ~permitted:false ~inherited:false
+end
+
+(* doesn't work if file has spaces but this is just for perf & it's good enough *)
+let split_init_rest_capabilities (input_text : string) =
+  match String.split input_text ~on:' ' with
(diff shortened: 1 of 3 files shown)


src/capability.ml
```diff
@@ -0,0 +1,216 @@
...
+      if linux_version_has_cap_perfmon linux_version
+      then
+        if does_not_remove_effective caps "cap_perfmon"
+        then true
+        else does_not_remove_effective caps "cap_sys_admin"
+      else does_not_remove_effective caps "cap_sys_admin"
+    else has_eff caps "cap_perfmon" || has_eff caps "cap_sys_admin")
+;;
+
+let run_test test_string linux_version =
```
Xyene, line 147: Let's move this into the test module, and name at least `~linux_version` (otherwise it's too easy to accidentally pass `test_string` in place of `linux_version`).

src/capability.ml
```diff
@@ -0,0 +1,216 @@
...
+  let split_version = String.split linux_version ~on:'.' in
+  let head = Int.of_string (List.hd_exn split_version) in
+  (Int.equal head 5 && Int.of_string (List.nth_exn split_version 1) >= 8) || head > 5
+;;
+
+let does_not_remove_effective t cap_string =
+  if contains_cap t cap_string then has_eff t cap_string else true
+;;
+
+let check_perf_support getcap linux_version =
```
Xyene, line 126: I think this function should probably live in `perf_capabilities.ml`, within `supports_tracing_kernel`.

src/capability.ml
```diff
@@ -0,0 +1,216 @@
...
+  List.mem (Map.keys t.caps_by_tag) cap_string ~equal:String.equal
+;;
+
+let get_cap_sets t cap_string =
+  if contains_cap t cap_string
+  then Map.find t.caps_by_tag cap_string |> Option.value ~default:t.all_caps
+  else Capability.empty
+;;
+
+let has_eff t cap_string = get_cap_sets t cap_string |> Capability.effective
```
Xyene, line 114:
```suggestion
let has_effective t cap_string = get_cap_sets t cap_string |> Capability.effective
```

Nit: avoid shortened function/variable names for readability.

PR #265, opened 2022-11-14
Fix dlfilter filtering out trace errors

Xyene:


cgaebel:
I just tested this and can confirm that dlfilter is no longer dropping trace errors. LGTM.

cgaebel (approved):

PR #277, opened 2023-07-10
Only check for fzf if required to find symbol

### LINKED ISSUE #276: fzf is required unnecessarily if -target is specified by symbol or default
With the program `int main() {}`:
```sh
$ _build/default/bin/magic_trace_bin.exe run -trigger main ./a.out
magic-trace could show you a fuzzy-finding selector here if "fzf" were in your PATH, but it is not.
```

And with
```c++
extern "C" [[gnu::noinline]] void magic_trace_stop_indicator() {}
int main() { magic_trace_stop_indicator(); }
```
```sh
$ _build/default/bin/magic_trace_bin.exe run -trigger . ./a.out
magic-trace could show you a fuzzy-finding selector here if "fzf" were in your PATH, but it is not.
```

Should be an easy fix.

ecatmur:
not if symbol is specified by name or defaulted to magic_trace_stop_indicator

Fixes #276.

### DIFF
--- a/core/symbol_selection.ml
+++ b/core/symbol_selection.ml
@@ -57,16 +57,16 @@ let evaluate ~supports_fzf ~elf ~header symbol_selection =
     | None -> Deferred.Or_error.error_string "No ELF found"
     | Some elf -> return elf
   in
-  let%bind () =
-    if force supports_fzf
-    then return ()
-    else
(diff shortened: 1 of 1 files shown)


Xyene (approved):
LGTM, thanks for the contribution!

PR #280, opened 2023-10-27
Fix stairstepping OCaml traces when exceptions are involved

Xyene:
This commit fixes two bugs:

* `caml_raise_exn` modifies the stack and then `ret`s into an `entertrap`, so we need to also do `entertrap` fixups on returns.

* Trap callstacks would often have the symbol of the parent frame re-inferred, which would result in doubling up of that frame. We fix this by seeding the trap callstack with a synthetic frame, which we make sure to delete before outputting trace events.

  A more principled approach would be to have `Callstack` know about traps directly and avoid creating new `Callstack.t`s for them, but that work is deferred to someday that isn't today.


cgaebel (approved):
lgtm

Issue #257, opened 2022-08-31
Exception raised while parsing perf output


dqminh:
magic-trace failed to parse perf output with the following errors:

```
(monitor.ml.Error
 ("BUG: exception raised while parsing perf output. Please report this to https://github.com/janestreet/magic-trace/issues/"
  (exn Not_found)
  (perf_output
   ("1907478/1909463 457407.880965552:          1                                branches:uH:   int                      564aa58813d4 Builtins_RunMicrotasks+0x554 (/usr/local/bin/workload) =>     564aa584fa00 Builtins_Call_ReceiverIsNotNullOrUndefined+0x0 (/usr/local/bin/workload)")))
 ("Raised at Base__Error.raise in file \"src/error.ml\" (inlined), line 9, characters 14-30"
  "Called from Base__Error.raise_s in file \"src/error.ml\", line 10, characters 19-40"
  "Called from Async_kernel__Pipe.map.(fun) in file \"src/pipe.ml\", line 1042, characters 65-70"
  "Called from Async_kernel__Job_queue.run_jobs in file \"src/job_queue.ml\", line 167, characters 6-47"
  "Caught by monitor Monitor.protect"))
```

I have:

```
perf --version
perf version 5.15.59
```
 
It looks to me that `int` flag is not recognized properly by magic-trace

cgaebel:
I ran this through our test suite and it looks like it's parsing `int` just fine. The part that's confusing it is the bit that says `1   ... branches:uH:`. I haven't seen that in my traces before.

Can you say more about how you invoked magic-trace and the machine you're running it on?

Also, can you take this again but with the environment variable setting `MAGIC_TRACE_DEBUG=1` and the extra argument `-z-print-perf-commands`, and send me the perf commands it output?

Sorry it took so long for us to get back to you. We're still here, just distracted with some other projects for a little while.

dqminh:
@cgaebel no problem at all, thanks for the reply 

> Can you say more about how you invoked magic-trace and the machine you're running it on?
Also, can you take this again but with the environment variable setting MAGIC_TRACE_DEBUG=1 and the extra argument -z-print-perf-commands, and send me the perf commands it output?

Here is the output of magic-trace.

```
perf record -o /tmp/magic_trace.tmp.0538ae/perf.data --timestamp --event=intel_pt/cyc=1,cyc_thresh=1,mtc_period=0/u -p 2993932 --snapshot=e
[ Attached. Press Ctrl-C to stop recording. ]
[...]

PR #288, opened 2024-01-08
Handle `tr end  async` trace events

Xyene:
Kernel commit

https://lore.kernel.org/lkml/20230928072953.19369-1-adrian.hunter@intel.com/

set `PERF_IP_FLAG_ASYNC` on trace end events. As per

https://github.com/torvalds/linux/blob/7ee022567bf9e2e0b3cd92461a2f4986ecc99673/tools/perf/builtin-script.c#L1546

I think this means that where `perf` would have previously outputted `tr end  jmp`, it now outputs `tr end  async`.

Adjust our regex to allow `async`, and treat it as a branch.

### DIFF
--- a/core/event.ml
+++ b/core/event.ml
@@ -2,6 +2,7 @@ open! Core
 
 module Kind = struct
   type t =
+    | Async
     | Call
     | Return
     | Syscall
--- a/core/event.mli
+++ b/core/event.mli
@@ -2,6 +2,7 @@ open! Core
 
 module Kind : sig
   type t =
+    | Async
     | Call
     | Return
     | Syscall
--- a/src/perf_decode.ml
+++ b/src/perf_decode.ml
@@ -22,7 +22,7 @@ let perf_callstack_entry_re = Re.Perl.re "^\t *([0-9a-f]+) (.*)$" |> Re.compile
 
 let perf_branches_event_re =
   Re.Perl.re
-    {|^ *(call|return|tr strt|syscall|sysret|hw int|iret|tr end|tr strt tr end|tr end  (?:call|return|syscall|sysret|iret)|jmp|jcc) +([0-9a-f]+) (.*) => +([0-9a-f]+) (.*)$|}
+    {|^ *(call|return|tr strt|syscall|sysret|hw int|iret|tr end|tr strt tr end|tr end  (?:async|call|return|syscall|sysret|iret)|jmp|jcc) +([0-9a-f]+) (.*) => +([0-9a-f]+) (.*)$|}
   |> Re.compile
 ;;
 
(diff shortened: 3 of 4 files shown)


Xyene:
@pikrzysztof, @wbthomason: this should fix https://github.com/janestreet/magic-trace/issues/257#issuecomment-1872989796, could you please give it a try?

You can get a prebuild binary from the GitHub action run: https://github.com/janestreet/magic-trace/actions/runs/7449992981

pikrzysztof:
Thanks for a quick contribution! This still crashes for me but with a different error

```
~/dev/web/org/queues/utilization-latency utilization-latency 17:11:34 RC=1» MAGIC_TRACE_DEBUG=1 ./magic-trace run -z-print-perf-commands ./target/debug/utilization-latency
perf record -o /tmp/magic_trace.tmp.c503dc/perf.data --timestamp --event=intel_pt/cyc=1,cyc_thresh=1,mtc_period=0/u --per-thread -t 120457 --snapshot=e
[ Snapshot taken. ]
[ perf record: Captured and wrote 0.128 MB /tmp/magic_trace.tmp.c503dc/perf.data ]
[ Finished recording. ]
perf script -i /tmp/magic_trace.tmp.c503dc/perf.data --ns --itrace=bep -F pid,tid,time,flags,ip,addr,sym,symoff,synth,dso,event,period --dlfilter /tmp/magic_trace.tmp.c503dc/perf_dlfilter.so
(monitor.ml.Error
 ("BUG: magic-trace devs thought this event was impossible, but you just proved them wrong. Please report this to https://github.com/janestreet/magic-trace/issues/"
  (event
   (Ok
    ((thread ((pid (120457)) (tid (120457)))) (time 7h29m46.678855932s)
     (data
      (Trace (trace_state_change End) (kind Jump) (src 0x562d04f3a2f0)
       (dst 0x0)))))))
  "Called from Magic_trace_lib__Trace_writer.write_event in file \"src/trace_writer.ml\", line 1035, characters 10-285"
  "Called from Async_kernel__Pipe.iter_without_pushback.(fun).loop in file \"src/pipe.ml\", line 894, characters 10-13"
  "Called from Async_kernel__Job_queue.run_jobs in file \"src/job_queue.ml\", line 167, characters 6-47"
```

Xyene:
Thanks for testing. I've pushed some changes that might resolve this--unfortunately I don't have an appropriate version of `perf` to test with, so there might be lingering issues.

Xyene:
Thanks for testing! Your trace has some overflow packets recorded, those map to data being lost due to the system under test running out of memory bandwidth.

cgaebel (approved):

PR #292, opened 2024-02-09
make it possible to invoke `magic-trace run -- prog args`

v-gb:
instead of `magic-trace run prog -- args`.

The current CLI is annoying because you can't just prefix the command line with `magic-trace run`, you have to insert of `--` in between the prog and the args.

I didn't actually run the result, because I get a number of type errors. I fixed enough to know my code types but not everything. Example of what I had to fix:
```diff
diff --git a/src/for_range.ml b/src/for_range.ml
index 5ddf0b0fff...4181efe477 100644
--- a/src/for_range.ml
+++ b/src/for_range.ml
@@ -21,7 +21,7 @@
 let range_hit_times ~decode_events ~range_symbols =
   let open Deferred.Or_error.Let_syntax in
   let%bind { Decode_result.events; _ } = decode_events () in
-  Deferred.List.map events ~f:(fun events ->
+  Deferred.List.map events ~how:`Sequential ~f:(fun events ->
     let { Trace_filter.start_symbol; stop_symbol } = range_symbols in
     let is_start symbol = String.(Symbol.display_name symbol = start_symbol) in
     let is_stop symbol = String.(Symbol.display_name symbol = stop_symbol) in
```

### DIFF
--- a/src/trace.ml
+++ b/src/trace.ml
@@ -575,7 +575,7 @@ module Make_commands (Backend : Backend_intf.S) = struct
       ~readme:(fun () ->
         "=== examples ===\n\n\
          # Run a process, snapshotting at ^C or exit\n\
-         magic-trace run ./program -- arg1 arg2\n\n\
+         magic-trace run -- ./program arg1 arg2\n\n\
          # Run and trace all threads of a process, not just the main one, snapshotting \
          at ^C or exit\n\
          magic-trace run -multi-thread ./program -- arg1 arg2\n\n\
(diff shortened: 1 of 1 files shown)


Xyene (approved):
Looks good to me, thanks! (And hi!)

We're still on core 0.15 which I think is what's causing your compilation errors. [If you have a switch for ocaml 4.13](https://github.com/janestreet/magic-trace/wiki/Build-instructions) I think that forces the constraints to resolve to core 0.15, which will get this repo building.

I think we should upgrade though, I'll take a look at doing that.

Xyene:
https://github.com/janestreet/magic-trace/pull/294

PR #298, opened 2024-04-30
Add support for jcmd Compiler.perfmap

theothornhill:
Even though the perf "spec" for jit languages define the perf map syntax, the jcmd Compiler.perfmap insists on prepending 0x to the addresses.

In addition, there's now a way for the jvm to generate the perfmap automatically on exit, making `magic-trace run -multi-thread -snapshot-size 1M -- ./demo/java` return something useful without messing around with perfmap generation. Add these options to the ./demo/java file

### DIFF
--- a/.ocamlformat
+++ b/.ocamlformat
@@ -1,2 +1,2 @@
-version=0.26.1
+version=0.26.2
 profile=janestreet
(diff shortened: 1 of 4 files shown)


theothornhill:
Ping! 😊

Xyene (approved):
LGTM, thanks for the contribution!

theothornhill:
Thanks - I'll look at the formatting failure. Code was run using the examples from the readme. I'll check if that either needs amending or if I missed something.

Xyene:
I've updated the docs, thanks!

PR #301, opened 2024-06-03
Reduce noise when OCaml exception tracking gets confused

Xyene:
We're not going to be able to address all the corner cases any time soon, so in the meantime output exception tracking warnings at most once per trace.


cgaebel (approved):
lmao

PR #306, opened 2024-10-04
Add support for compressed output files, whether gzip (supported by Pe…

### LINKED ISSUE #154: Compressed trace files
Trace files are big and compress really well, 5x-ish in my experience. Could magic-trace write out compressed traces instead of raw traces?

Perfetto already has support for loading gzipped traces. But gzip is really slow (fingers crossed for Sapphire Rapids, but that's years from being commonplace).

As far as I can tell, we have two options here:

- teach magic-trace to write gzipped traces, or
- teach magic-trace to write `zstd`'d traces, and also teach perfetto to load them.

The first option is easier, but the second option would be faster to save and load. FWIW, Tudor thinks zstd is overkill because we're so slow at decoding otherwise.

- [Perfetto's gzip_trace_parser.cc](https://github.com/janestreet/perfetto/blob/jane-public-patches/src/trace_processor/importers/gzip/gzip_trace_parser.cc)
- [zstd ocaml bindings](https://ocaml.janestreet.com/ocaml-core/v0.12/doc/zstandard/Zstandard/index.html)
- [gzip ocaml bindings](https://github.com/hcarty/ezgzip)
- [libdeflate](https://github.com/ebiggers/libdeflate) has the tall claim of gzipping faster that zstd

bnigito:
…rfetto) or zstd (not yet supported)

This builds on previous work for compressed file destinations

Closes #154

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,10 +37,31 @@ jobs:
           wget http://ftp.debian.org/debian/pool/main/u/upx-ucl/upx-ucl_3.96-2_amd64.deb
           sudo dpkg -i upx-ucl_3.96-2_amd64.deb
 
-      - name: Use OCaml ${{ matrix.ocaml-version }}
+      - name: "Install apt packages"
         run: |
           sudo apt-get update
           sudo apt-get install bubblewrap musl-tools
+
+      - name: Build zlib with musl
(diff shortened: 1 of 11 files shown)


Xyene (approved):
LGTM, thanks!

PR #308, opened 2024-10-15
Handle tsx transactions

bnigito:
Add limited support for handling Intel's TSX (Transactional Synchronization Extensions), specifically Restricted Transactional Memory (RTM) where aborted transactions were not properly handled.

This also makes a minor change which corrects error printing, where a flush was intended (%!) but the % accidentally omitted.


Xyene (approved):
LGTM, thanks!

PR #320, opened 2025-01-16
Snapshot using perf ctlfd

lnbrks:


src/perf_ctlfd.ml
```diff
@@ -0,0 +1,80 @@
...
+  let ctl_rx, ctl_tx = Core_unix.pipe ~close_on_exec:false () in
+  let ack_rx, ack_tx = Core_unix.pipe ~close_on_exec:false () in
+  Core_unix.set_close_on_exec ctl_tx;
+  Core_unix.set_close_on_exec ack_rx;
+  { ctl_rx; ctl_tx; ack_rx; ack_tx; ack_buf = Bytes.make (String.length ack_msg) '\000' }
+;;
+
+let close_perf_side_fds t =
+  Core_unix.close ~restart:true t.ctl_rx;
+  t.ctl_rx <- Core_unix.File_descr.of_int (-1);
```
Xyene, line 30: Nit: let's make these `option`s.

src/perf_tool_backend.ml
```diff
@@ -412,35 +447,35 @@ module Recording = struct
...
-           | `sigusr2 -> Signal.usr2)
+      | At_exit _, `ctrl_c -> true
     in
-    match signal with
-    | None -> ()
-    | Some signal -> Signal_unix.send_i signal (`Pid t.pid)
+    if should_take_snapshot
+    then (
+      match t.snapshot_behavior with
+      | Never -> failwith "unreachable"
```
Xyene, line 473: Suggestion: lift this block into a `let snapshot () = ...`, then have the `-> (true|false)` branches above be `-> (() | snapshot ())`. That way we don't have to double-state the `Never` case.

src/perf_ctlfd.ml
```diff
@@ -0,0 +1,80 @@
...
+      !total_bytes_read < Bytes.length t.ack_buf
+    | _ -> failwith "unreachable"
+  do
+    ()
+  done;
+  if not
+       (String.equal
+          ack_msg
+          (Bytes.unsafe_to_string ~no_mutation_while_string_reachable:t.ack_buf))
+  then failwith "Receive malformed ack from perf";
```
Xyene, line 69: This will be very hard to debug if it ever raises. Let's turn this into a `raise_s` and include the ack we did get.

src/perf_tool_backend.ml
```diff
@@ -353,23 +387,23 @@ module Recording = struct
...
+    let snapshot_behavior : Snapshot_behavior.t =
       if full_execution
-      then `never
+      then Never
       else (
         match when_to_snapshot with
         | Magic_trace_or_the_application_terminates ->
-          if perf_supports_snapshot_on_exit then `at_exit `sigint else `at_exit `sigusr2
-        | Application_calls_a_function _ -> `function_call)
+          if perf_supports_snapshot_on_exit then At_exit Sigint else At_exit Sigusr2
```
Xyene, line 396: Can we use the ctrlfd here too?

This logic is also used for triggering on ^C; it'd be nice to get away from the buggy signal handling where we can.

PR #332, opened 2025-07-15
Improve kernel tracing capability detection

ipenas-cl:
This commit enhances the kernel tracing capability check to be more accurate and less restrictive. Instead of only checking if the user is root, it now checks multiple conditions:

1. Root user (traditional method)
2. perf_event_paranoid setting (< 0 allows kernel tracing)
3. CAP_SYS_ADMIN capability on the perf binary
4. Fallback to conservative check if other methods fail

This addresses issue #176 and allows non-root users to use kernel tracing when properly configured (e.g., via perf_event_paranoid=-1 or capabilities on the perf binary).

The implementation follows the guidance from:
https://perf.wiki.kernel.org/index.php/Perf_tools_support_for_Intel%C2%AE_Processor_Trace#Adding_capabilities_to_perf

### DIFF
--- a/src/perf_capabilities.ml
+++ b/src/perf_capabilities.ml
@@ -79,13 +79,51 @@ let supports_last_branch_record () =
 ;;
 
 let supports_tracing_kernel () =
-  (* Only allow tracing the kernel if we are root. `perf` will start even without this,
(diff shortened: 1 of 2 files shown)


dkalinichenko-js:
Closing: low quality, AI.

PR #353, opened 2026-02-09
Trace segments alpha + inlined frames

Svetlitski:
# Introduce new, experimental trace-writing backend

The existing `Trace_writer` module works well enough (albeit not perfectly) most of the time. However, it is difficult to reason about, in large part because it writes the trace in a streaming fashion. That introduces significant additional complexity and bookkeeping, and limits the ability of the trace-writer to make use of information discovered later in the trace (I believe the latter is why traces produced today often have the few frames closest to the root wrong). Because we want to extend the trace-writer with new functionality, we're starting fresh with a different design that's easier to reason about. The new implementation currently exists alongside the original, but the goal is to eventually replace it entirely.

Instead of writing the trace in a streaming fashion, we construct an internal representation of the trace in memory, and write out the trace in a separate, final pass once all of the events have been consumed. The module responsible for doing most of the heavy lifting is the new `Trace_segment`, which represents a continuous, lossless, and error-free segment of the trace; we create a new trace-segment whenever we encounter an error.

The other major addition is that the new implementation includes inlined function calls, using LLVM for symbolization, dramatically increasing the fidelity of the trace.

**This PR is effectively an alpha of the new implementation.** The code here does indeed work,
and produces better traces than the existing backend in many cases, but there are a couple
critical issues:
1. **Error recovery**: We create a new trace-segment whenever we encounter an error,
   **but at present we naively treat each trace-segment as disjoint**. We need to add an additional
   "stitching" pass before the trace is written out, making a heuristic, best-effort attempt to
   join together adjacent trace-segments in a way that preserves control-flow continuity.
   All of this is a long way of saying that if you encounter *any* error while using the new
   implementation, your trace is likely to be horribly broken.
2. **Performance**: Including inlined frames makes the traces significantly larger, and the supporting
   code for this new functionality is written pretty naively from a performance standpoint. As a result,
   the new implementation is roughly 2x slower than the old implementation.

It should also go without saying that while this code appears to work well on the traces I've tried it on, I would not at all be surprised if there are still bugs/edge-cases.

We will address these shortcomings over time, but in the meantime the new implementation is opt-in; setting the environment variable `MAGIC_TRACE_USE_NEW_TRACE_WRITER=1` will enable it.


src/trace_segment.mli
```diff
@@ -0,0 +1,25 @@
...
+
+val in_filtered_region : t -> bool
+val add_event : t -> Event.Ok.Data.t -> Timestamp.t -> unit
+
+val write_trace
+  :  t
+  -> (module Trace_writer_intf.S_trace with type thread = 'thread)
+  -> 'thread
+  -> Elf.Addr_table.t
+  -> enter_initial_callstack:bool
```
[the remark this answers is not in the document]
Xyene: ...actually, what is the intent of this? I only see this being called with `true` for both params.

src/trace_segment.ml
```diff
@@ -0,0 +1,1080 @@
...
+  ; ocaml_exception_info : Ocaml_exception_info.t Or_null.t
+  ; exception_handlers : Frame.t Vec.t
+  (** The currently active OCaml exception handlers. This is used to determine which frame
+      to return to when [ocaml_exception_info] indicates that the current event is an
+      OCaml exception being raised in the traced program.
+
+      In contrast to [callstacks] — which records the entire history of control-flow for
+      later examination — [exception_handlers] represents the state **as of the event we
+      are currently processing**, and as such is only used during the "ingestion" phase
+      (i.e. while calls are still being made to [add_event]). *)
```
Xyene: I'm assuming that an invariant here is that every `Frame.t` in `exception_handlers` is present in `hd callstacks`. Is that right?

src/trace_segment.ml
```diff
@@ -0,0 +1,1080 @@
...
+   is that in the context of a call, [src] is the parent frame of the call to [dst]
+   and thus *it continues to exist*. We want our callstacks to reflect that. *)
+let handle_call (t : t) (time : Timestamp.t) ~(src : Location.t) ~(dst : Location.t) =
+  (* First, reconcile things such that [src] matches [current_frame t] if it doesn't
+     already. *)
+  let () =
+    match Frame.find (current_frame t) src.symbol with
+    | #(This _, ~distance:0) -> (* The happy case, [src] matches [current_frame t]. *) ()
+    | #(This src_frame, ~distance) ->
+      (* [src] exists, but is higher up the callstack. *)
```
Xyene: When can this happen?

src/nonempty_vec.mli
```diff
@@ -0,0 +1,19 @@
+open! Core
+
+module type%template [@kind k = (value, value & value & value)] S := sig
+  (** A [Vec.t] guaranteed to contain at least one element. *)
+  type ('a : k) t
+
+  val create : 'a -> 'a t
+  val length : _ t -> int
```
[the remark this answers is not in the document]
Svetlitski: Yes, because it does not depend on `'a`.

src/trace_segment.mli
```diff
@@ -0,0 +1,25 @@
+open! Core
+
+(** A continuous, lossless, error-free segment of a trace corresponding to a single
+    thread. *)
+type t
+
+val create : Ocaml_exception_info.t option -> in_filtered_region:bool -> t
```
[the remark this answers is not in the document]
Svetlitski: Even once it's the only argument (as I'm getting rid of `in_filtered_region` in favor of externalizing that in response to your other review comments)? I'm happy either way, just clarifying.

src/trace_writer_implementation_intf.ml
```diff
@@ -0,0 +1,74 @@
...
+  end
+
+  val create_expert
+    :  trace_scope:Trace_scope.t
+    -> debug_info:Elf.Addr_table.t option
+    -> ocaml_exception_info:Ocaml_exception_info.t option
+    -> earliest_time:Time_ns.Span.t
+    -> hits:(string * Breakpoint.Hit.t) list
+    -> annotate_inferred_start_times:bool
+    -> (module Trace with type thread = _)
```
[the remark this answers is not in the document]
Svetlitski: This is not my addition, it is present [in the original code](https://github.com/janestreet/magic-trace/blob/d31ceba65eee953a1bb8c7273f5a34f80c136ebd/src/trace_writer.mli#L59). Removing it results in a compilation error.

src/trace_segment.mli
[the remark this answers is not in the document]
Svetlitski: Yes, at present `enter_initial_callstack` and `exit_final_callstack` are always `true`. The intent is that once we implement support for "stitching" these will become useful. I'm happy to delete these arguments until we actually have stitching support if you'd prefer.

src/trace_segment.ml
```diff
@@ -0,0 +1,1080 @@
...
+    }
+
+  val create : Location.t -> parent:t -> t
+
+  (** Find the first frame whose [location.symbol] matches the provided argument.
+
+      Returns the matching frame (if found), and that frame's distance from the initial
+      frame (e.g. a call to [find my_frame my_symbol] with a return value of
+      [#(This _, ~distance:0)] indicates that [my_frame.location.symbol] is [my_symbol]). *)
+  val find : t -> Symbol.t -> #(t Or_null.t * distance:int)
```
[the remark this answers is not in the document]
Svetlitski: I agree the latter type would be more expressive, were it not invalid; you can only have an `('a : value) or_null`, which precludes product layouts.

src/trace_segment.ml
```diff
@@ -0,0 +1,1080 @@
...
+      | { parent = This parent; _ } -> find_ancestor parent ~ancestor (distance + 1))
+  ;;
+
+  let find_ancestor t ~ancestor = find_ancestor t ~ancestor 0
+
+  module Sentinel = struct
+    type nonrec t = t
+
+    let sentinel_location : Location.t =
+      { instruction_pointer = 0L; symbol_offset = 0; symbol = From_perf "\x00" }
```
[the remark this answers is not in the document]
Svetlitski: For the sake of being defensive, we define a special symbol for the sentinel which is *impossible* to confuse with a real symbol, because ELF symbol names are NULL-terminated strings. Upon thinking about this more, this should probably be `From_perf "\x00_"` to actually guarantee that. See https://refspecs.linuxbase.org/elf/gabi4+/ch4.strtab.html . I'll add a comment explaining this.

src/trace_segment.ml
```diff
@@ -0,0 +1,1080 @@
...
+    (** [distance] indicates how many frames this return pops off of the callstack.
+        [distance = 1] is the usual case of returning from the current frame to its
+        parent. *)
+end
+
+module Callstack = struct
+  type t =
+    #{ time : Timestamp.t
+     ; leaf : Frame.t
+     ; control_flow : Control_flow.t
```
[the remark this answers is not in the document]
Svetlitski: See the comment on `callstacks` for an explanation of this.

src/trace_segment.ml
Svetlitski: Yes.

src/trace_segment.ml
```diff
@@ -0,0 +1,1080 @@
...
+  (** The currently active OCaml exception handlers. This is used to determine which frame
+      to return to when [ocaml_exception_info] indicates that the current event is an
+      OCaml exception being raised in the traced program.
+
+      In contrast to [callstacks] — which records the entire history of control-flow for
+      later examination — [exception_handlers] represents the state **as of the event we
+      are currently processing**, and as such is only used during the "ingestion" phase
+      (i.e. while calls are still being made to [add_event]). *)
+  ; mutable last_known_instruction_pointer : int64
+  ; in_filtered_region : bool
```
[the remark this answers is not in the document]
Svetlitski: I agree, I meant to extract this and forgot to do so.

src/trace_segment.ml
```diff
@@ -0,0 +1,1080 @@
...
+  { existing with
+    callstacks =
+      Nonempty_vec.create
+        (#{ last_callstack with control_flow = Return { distance = 0 } } : Callstack.t)
+  ; exception_handlers = Vec.copy existing.exception_handlers
+  ; in_filtered_region
+  }
+;;
+
+let in_filtered_region t = t.in_filtered_region
```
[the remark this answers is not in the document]
Svetlitski: We won't have this field anymore period, see above comment.

src/trace_segment.ml
```diff
@@ -0,0 +1,1080 @@
...
+
+        Using a sentinel allows us to avoid a variety of special-cases, and lets us update
+        the root of all callstacks in a trace in O(1) time. *)
+    type t = private frame
+
+    val create : unit -> t
+
+    (** Mutate [t]'s contents to the provided [location] and [parent] and return [t] as a
+        [frame]. *)
+    val become_frame : t -> Location.t -> parent:frame -> frame
```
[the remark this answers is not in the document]
Svetlitski: Right now? No. Once we implement stitching? Yes.

src/trace_segment.ml
```diff
@@ -0,0 +1,1080 @@
...
+;;
+
+(* [handle_call] uses [src], unlike the other event handlers. The rationale for this
+   is that in the context of a call, [src] is the parent frame of the call to [dst]
+   and thus *it continues to exist*. We want our callstacks to reflect that. *)
+let handle_call (t : t) (time : Timestamp.t) ~(src : Location.t) ~(dst : Location.t) =
+  (* First, reconcile things such that [src] matches [current_frame t] if it doesn't
+     already. *)
+  let () =
+    match Frame.find (current_frame t) src.symbol with
```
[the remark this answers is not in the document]
Svetlitski: It's been ~1 month since I touched this, but IIRC I ran numerous experiments and I recall I _never_ hit most of these codepaths.

src/trace_segment.ml
```diff
@@ -0,0 +1,1080 @@
...
+  assert (Timestamp.( >= ) time t.last_event_time);
+  t.last_event_time <- time;
+  (match t.ocaml_exception_info with
+   | Null -> ()
+   | This ocaml_exception_info ->
+     (match event with
+      | Trace { src; dst; _ } ->
+        Ocaml_exception_info.iter_pushtraps_and_poptraps_in_range
+          ocaml_exception_info
+          ~from:t.last_known_instruction_pointer
```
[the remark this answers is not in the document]
Svetlitski: Yes, this works because `Ocaml_exception_info.iter_pushtraps_and_poptraps_in_range` is binary-searching under-the-hood.

src/new_trace_writer.ml
```diff
@@ -0,0 +1,971 @@
+open! Core
+module Nonempty_vec = Nonempty_vec.Value
+
+let debug = ref false
+let is_kernel_address addr = Int64.(addr < 0L)
```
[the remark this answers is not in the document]
Svetlitski: I am somewhat disinclined to do this because rending them apart made it possible for me to understand what remains to be replaced, which I found very difficult before.

src/trace_writer_implementation_intf.ml
```diff
@@ -0,0 +1,74 @@
...
+      ; callstack : Callstack_compression.compression_event
+      }
+    [@@deriving sexp, bin_io]
+  end
+
+  val create_expert
+    :  trace_scope:Trace_scope.t
+    -> debug_info:Elf.Addr_table.t option
+    -> ocaml_exception_info:Ocaml_exception_info.t option
+    -> earliest_time:Time_ns.Span.t
```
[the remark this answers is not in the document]
Svetlitski: That abstraction is currently internal to trace-writer.

src/interned_string.ml
```diff
@@ -0,0 +1,23 @@
+open! Core
+
+type t = string [@@deriving sexp_of]
+
+let equal = phys_equal
+let compare t1 t2 = Int.compare (Obj.magic t1) (Obj.magic t2)
+let hash t = Int.hash (Obj.magic t)
```
Svetlitski, line 7: This is problematic because the hash will change if the string is relocated, either via promotion to the major heap or during compaction.

src/new_trace_writer.ml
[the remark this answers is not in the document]
Svetlitski: Thinking about this more, I think we could delete more of this code if I modify things such that we always use the old trace-writer when we are processing LBR, regardless of what `MAGIC_TRACE_USE_NEW_TRACE_WRITER` is set to. This would be temporary until the new trace-writer has its own LBR support, of course. Thoughts @Xyene ?

src/nonempty_vec.mli
```diff
@@ -0,0 +1,20 @@
...
+  val first : 'a t -> 'a
+  val get : 'a t -> int -> 'a
+  val set : 'a t -> int -> 'a -> unit
+  val last : 'a t -> 'a
+  val push_back : 'a t -> 'a -> unit
+  val iter : 'a t -> f:local_ ('a -> unit) -> unit
+  val iter_pairs : 'a t -> f:local_ (#('a * 'a) -> unit) -> unit
+end
+
+module Value : S [@kind value]
```
Svetlitski: This instantiation is unused, I'll remove it.

src/perf_decode.ml
```diff
@@ -109,9 +112,12 @@ let parse_event_header line =
           "Regex of perf output did not match expected fields" (results : string array)])
 ;;
 
-let parse_symbol_and_offset ?perf_maps pid str ~addr : Symbol.t * int =
-  match Re.Group.all (Re.exec symbol_and_offset_re str) with
-  | [| _; symbol; offset |] ->
+let parse_symbol_and_offset_and_dso ?perf_maps pid str ~addr
+  : Symbol.t * int * Interned_string.t
```
Svetlitski: Make this an unboxed tuple.

src/slice.ml
```diff
@@ -0,0 +1,36 @@
+open! Core
```
Svetlitski, line 1: I should check if `Ox` already has a version of this.

src/symbolizer.ml
```diff
@@ -0,0 +1,103 @@
...
+  (** This is ordered root-to-leaf such that the entry at index 0 is the physical frame,
+      and the subsequent entries are the inlined frames. *)
+  type t = Info.t iarray [@@deriving sexp_of, equal, hash, compare]
+
+  let physical_frame t = Iarray.unsafe_get t 0
+  let inlined_frames t = Slice.create t ~pos:1 ~len:(Iarray.length t - 1)
+end
+
+module Llvm_symbolizer = struct
+  type t = Iptr.t
```
Svetlitski:
```suggestion
  type t = private Iptr.t
```

src/interned_string.mli
```diff
@@ -0,0 +1,6 @@
+open! Core
```
Svetlitski, line 1: Given that this module is only used to intern DSO/executable paths, I think it makes more sense to replace this with an equivalent `Build_id` module, where instead of `Interned_string.intern` you have `Build_id.resolve` which gets the `Build_id` corresponding to that path, resolved via LLVM.

src/symbolizer.ml
```diff
@@ -0,0 +1,103 @@
...
+  external symbolize
+    :  t
+    -> executable:Interned_string.t
+    -> addr:i64
+    -> Response.t or_null
+    = "caml_no_bytecode_impl" "magic_trace_llvm_symbolize_address"
+end
+
+type t =
+  { symbolization_cache : (Request.t, Response.t Uopt.t) Hashtbl.t
```
Svetlitski: A `(value, value_or_null)` template on `Hashtbl.t` made it into the latest public-release of `Core`, so we should use that instead of this `Uopt.t` workaround.

src/symbolizer.mli
```diff
@@ -0,0 +1,57 @@
...
+  val display_name : t -> string
+
+  (** This is currently a gross hack, to be used solely for inlined frames. *)
+  val to_location : t -> Event.Location.t
+end
+
+module Response : sig
+  type t [@@deriving sexp_of]
+
+  val physical_frame : t -> Info.t
```
Svetlitski, line 14: Yes, this isn't used at present. The idea is it will be useful when we completely replace Owee with LLVM.

.gitignore
```diff
@@ -4,3 +4,7 @@ _build
 _opam
 .vscode/*
 !.vscode/tasks.json
+*.fxt
```
Svetlitski, line 9: Add `perf.data`, `perf.data.old`

dune-project
```diff
@@ -1 +1 @@
-(lang dune 2.0)
+(lang dune 3.0)
```
Svetlitski, line 1: I needed to change this in order to get access to the `link_flags` environment setting in the top-level `dune`, which we need for our statically linked build to work. This came with the unfortunate side-effect of changing `dune`'s auto-formatting of its own files, hence why most of the project's `dune` files are modified in this PR.

Svetlitski (commented):
Note that this review does _not_ cover `Trace_segment`; I'll take a final pass over that module on Monday.

src/trace_segment.ml
```diff
@@ -0,0 +1,1552 @@
...
+          parent
+    ;;
+
+    let to_string_list t = to_string_list [] t
+    let print_callstack leaf = to_string_list leaf |> String.concat_lines |> printf "%s"
+  end
+end
+
+module Control_flow = struct
+  type t =
```
Svetlitski, line 339: Making this `private` and gating construction behind `val call : int -> t` and `val return : int -> t` functions that internally return values from a statically allocated array when `0 <= {depth,distance} <= 15` would be an easy way to eliminate most allocations for this type.

src/trace_segment.ml
```diff
@@ -0,0 +1,1552 @@
...
+    [%test_eq: Interned_string.t] current_physical_frame.location.dso dso);
+  diff_inlined_frames'
+    t
+    time
+    ~dso
+    ~before:(symbolize_inlined_frames t ~dso ~addr:before)
+    ~after:(symbolize_inlined_frames t ~dso ~addr:after)
+;;
+
+let append_inlined_frames t time ~(physical_frame : Frame.t) ~physical_frame_is_new =
```
Svetlitski: This deserves a doc-comment explaining why it's important to introduce the `physical_frame` and all of its inlined children in a single `Call { ... }` event.

src/trace_segment.ml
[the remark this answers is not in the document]
Svetlitski: I reordered this.

src/trace_segment.ml
```diff
@@ -0,0 +1,1080 @@
...
+    (* We start our search for [dst] from the parent of the current frame because
+       otherwise you'd incorrectly handle non-tail recursion, and because returning to the
+       current frame is impossible anyway. We add 1 to [distance] in the [control_flow] to
+       account for the one extra frame implicitly traversed by doing this. *)
+    (match Frame.find parent_frame dst.symbol with
+     | #(This dst_frame, ~distance) ->
+       (* 99% of the time [distance] should be 0, indicating we are returning to
+          [parent_frame] as expected. We allow for the possibility of "long" returns to
+          account for [Sysret]/[Iret] events that return to userspace directly from deep
+          within their kernel/interrupt stack. *)
```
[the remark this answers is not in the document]
Svetlitski: I believe the answer is "no", for much the same reason that correctly handling OCaml exceptions requires additional information from debuginfo (or specialized introspection of the control-flow).

src/trace_segment.ml
```diff
@@ -0,0 +1,1552 @@
...
+  else (
+    match current_physical_frame.parent with
+    | Null ->
+      (* This is probably a non-recursive tail-call, but we don't know anything
+         about the previous frame, so we treat this is a [Call] because we only
+         want to emit a frame-enter while writing out the trace. *)
+      let dst_frame = Frame.create dst ~parent:(t.root :> Frame.t) ~kind:Physical in
+      append_inlined_frames t time ~physical_frame:dst_frame ~physical_frame_is_new:true
+    | This parent ->
+      (* This is probably a non-recursive tail-call. *)
```
Svetlitski, line 748: It's outside the scope of this PR, but in the future it would be nice to retain this information in the `Callstack.t` so that we can produce a [flow event](https://fuchsia.dev/fuchsia-src/development/tracing/trace_events#flow-event) to make the control-flow more obvious in the UI.

src/trace_segment.ml
```diff
@@ -0,0 +1,1552 @@
...
+         t
+         time
+         ~physical_frame:dst_frame
+           (* This is subtle; Yes the frame is new, but we *don't* want to reflect that in a
+              [Call], because discovered roots are handled separately. *)
+         ~physical_frame_is_new:false
+     | #(This dst_frame, ~distance, ~leaf_of_inlined_stack, ..) ->
+       log_unexpected_case
+         [%message
+           "[exception_handlers] appears to be out-of-sync with [callstacks]; there is \
```
Svetlitski: Upon further reflection, hitting this doesn't necessarily indicate that there's a bug. Imagine you have a function like:
```ocaml
let rec process_data x n = 
  match n with 
  | 100 -> x
  | 50 -> ( try (process_data x (n + 1)) with _ -> process_data x 76 )
  | 75 -> failwith "Raise an exception"
  | n -> n + process_data (x + n) (n + 1)
;;
```

If the program calls `process_data 0xf00 0`, and magic-trace starts tracing during the execution of `process_data _ 70`, then when the program eventually reaches the `75` case where it raises an exception, the symbol of the exception target `process_data` will be found in the current callstack, but `t.exception_handlers` will be empty because the pushtrap only occurs in the `50` case, which happened before we started tracing.

src/trace_segment.ml
```diff
@@ -0,0 +1,1552 @@
...
+   | _ -> ());
+  (match event with
+   (* TODO Get the untraced "kind" right instead of always showing [Location.untraced] for untraced time. *)
+   | Trace { trace_state_change = Some Start; dst; _ } -> handle_return t time ~dst
+   | Trace { trace_state_change = Some End; src; dst = _; _ } ->
+     handle_call t time ~src ~dst:Location.untraced
+   | Trace { trace_state_change = None; kind = Some kind; src; dst } ->
+     (match kind with
+      | Call | Syscall | Hardware_interrupt | Interrupt -> handle_call t time ~src ~dst
+      | (Return | Jump) when is_ocaml_exception_handler t ~dst ->
```
Svetlitski: The old trace-writer implementation also checks this for `Interrupt`, which I find somewhat surprising. Reading [the commit that introduced this in the old trace-writer](https://github.com/janestreet/magic-trace/commit/3e35d642b61ab1e75640234b1c4a2c7e3536e42f) isn't particularly enlightening. @Xyene do you have any additional context here?

src/trace_segment.ml
```diff
@@ -0,0 +1,1552 @@
...
+       let main () = fn1 ()
+       *)
+  let%expect_test "Tail-call" =
+    let #(~t, ~call, ~return, ~jump) = setup_test () in
+    call ~src:"main" ~dst:"fn1";
+    (* Tail-call [fn2] from [fn1] *)
+    jump ~src:"fn1" ~dst:"fn2";
+    return ~src:"fn2" ~dst:"main";
+    print_callstacks t;
+    (* TODO Showing a return to [main] and subsequent call into [fn2] is a bug that was
```
Svetlitski: Delete this comment.

Svetlitski (commented):
Took another pass at `Trace_segment`

src/trace_segment.ml
```diff
@@ -0,0 +1,1552 @@
...
+        && Int64.( <> ) t.last_known_location.instruction_pointer 0L
+        && Int64.( <> ) src.instruction_pointer 0L
+     then (
+       let mutable prev_inlined_frames =
+         symbolize_inlined_frames
+           t
+           ~dso:t.last_known_location.dso
+           ~addr:t.last_known_location.instruction_pointer
+       in
+       (* TODO I fear symbolizing at every byte like this is likely to be *very* expensive, but let's focus on just getting things working for now. *)
```
Svetlitski, line 932: According to Claude's reading of the LLVM source, it would not be hard to make a small patch to LLVM that surfaces the enclosing address range for a queried address, so we could avoid making `N` symbolization requests for `N` bytes of contiguous code that all have the same debuginfo. Given just how expensive this has turned out to be, this seems worthwhile.

src/trace_writer.ml
```diff
@@ -1231,3 +1233,5 @@ and write_event' (T t) ?events_writer event =
         | None, _ -> ());
        if !debug then print_s (sexp_of_inner t))
 ;;
+
+let finalize t = end_of_trace t
```
Xyene, line 1237: Nit: can we rename one of these, and not have the trampoline?

src/trace_segment.mli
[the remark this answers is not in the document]
Xyene: `TODO` or removal both sound fine to me. Maybe removal slightly better?

src/symbolizer.ml
```diff
@@ -0,0 +1,101 @@
...
+    ; executable : Interned_string.t
+    }
+  [@@deriving compare, sexp_of, hash]
+end
+
+module Info = struct
+  type t = { demangled_name : string }
+  [@@unboxed] [@@deriving equal, compare, hash, sexp_of]
+
+  let display_name { demangled_name } = demangled_name ^ " (inlined)"
```
Xyene: In an ideal world (likely as a `TODO`), this ought to be a single bit in the event we write out to Perfetto, and we patch Perfetto to display `(inlined)` here.

Otherwise, inlined vs. un-inlined frames will (1) take more trace space, and (2) get hash-colored differently (which feels confusing).

src/perf_decode.ml
```diff
@@ -109,9 +112,12 @@ let parse_event_header line =
           "Regex of perf output did not match expected fields" (results : string array)])
 ;;
 
-let parse_symbol_and_offset ?perf_maps pid str ~addr : Symbol.t * int =
-  match Re.Group.all (Re.exec symbol_and_offset_re str) with
-  | [| _; symbol; offset |] ->
+let parse_symbol_and_offset_and_dso ?perf_maps pid str ~addr
+  : #(Symbol.t * int * Interned_string.t)
```
Xyene: Hmm, if we're changing this type, wouldn't it be nicer if we did the string conversion at L142 upstream? We already have `Symbol.Unknown`; I think the only reason this code was written this way was to not pass up `dso` to the caller, but we're doing that anyway now...

Xyene: Semantically, isn't this `Interned_string.t or_null`?

src/event.ml
```diff
@@ -25,9 +25,11 @@ end
 
 module Location = struct
   type t =
-    { instruction_pointer : Int64.Hex.t
+    { (* TODO Use [i64] everywhere. *)
+      instruction_pointer : Int64.Hex.t
     ; symbol : Symbol.t
     ; symbol_offset : Int.Hex.t
+    ; dso : Interned_string.t
```
Xyene: I left this comment elsewhere too, but `Interned_string.t or_null` seems like the more natural representation here?

.github/workflows/build.yml
```diff
@@ -16,70 +16,31 @@ jobs:
...
-          filename='zlib-1.3.2.tar.gz'
-          wget "https://zlib.net/$filename"
-          shasum -a 256 --check <(echo "bb329a0a2cd0274d05519d61c667c062e06990d72e125ee2dfa8de64f0119d16  $filename")
-          tar -xzf "$filename" -C musl-zlib --strip-components=1
-          cd musl-zlib
-          CC=musl-gcc ./configure --libdir=/usr/lib/x86_64-linux-musl --includedir=/usr/include/x86_64-linux-musl
-          make -j$(nproc)
-          sudo make install
-
-      - name: Build zstd with musl
```
Xyene, line 64: Remind me, are we no longer building statically with musl?

.github/workflows/build.yml
```diff
@@ -97,11 +58,12 @@ jobs:
 
       - run: opam exec -- make PROFILE=static
 
-      - run: opam exec -- dune runtest
+      - run: opam exec -- dune runtest --profile=static
 
       - name: Compress magic-trace executable
         run: |
           cp _build/install/default/bin/magic-trace .
+          chmod u+w magic-trace
```
Xyene, line 66: OOC, why was this necessary?

src/new_trace_writer.ml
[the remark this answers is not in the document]
Xyene: SGTM

src/trace_segment.ml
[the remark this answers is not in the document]
Xyene: Ack, worth a comment.

src/trace_segment.ml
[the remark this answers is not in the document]
Xyene: Ack, worth a comment + `TODO`.

src/trace_segment.ml
Xyene: Ack, worth a comment.

.github/workflows/build.yml
Svetlitski: We are still building statically with `musl`. What's changed is that we're now using an Alpine Docker image, which is musl-native, so we can get musl-built libraries directly from the package manager.

src/symbolizer.ml
Svetlitski: Agreed. Even a ["Null argument"](https://fuchsia.dev/fuchsia-src/reference/tracing/trace-format#null-argument) takes up 8 bytes, which adds up fast when you have tons of inlined frames. I'll see if maybe I can steal a bit from the ["Category" field](https://fuchsia.dev/fuchsia-src/reference/tracing/trace-format#event-record) of the event header for this.

.github/workflows/build.yml
Svetlitski: I haven't the faintest clue TBH. I ran the workflow, it failed at this point because the executable wasn't writable, so I made it writable.

src/trace_segment.mli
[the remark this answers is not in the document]
Svetlitski: I can remove it, sure.

src/trace_segment.ml
```diff
@@ -0,0 +1,1080 @@
...
+
+  (** Iterate from leaf-to-root up to the given number of frames, or until encountering
+      the [Sentinel.t] *)
+  val iter_n : t -> int -> f:local_ (t -> unit) -> unit
+
+  (** Iterate the entire callstack from root-to-leaf. Note that this is *not*
+      tail-recursive, given that frames form a singly-linked list from leaf-to-root. *)
+  val iter_rev : t -> f:local_ (t -> unit) -> unit
+
+  val find_ancestor : t -> ancestor:t -> int Or_null.t
```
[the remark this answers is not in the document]
Svetlitski: I added a comment; this returns an unboxed tuple now with labeled fields in the spirit of your suggestion.

src/trace_segment.ml
Svetlitski: Added a comment and also added code to check the invariant under `MAGIC_TRACE_CHECK_INVARIANTS`.

src/trace_segment.ml
Svetlitski: I have a working prototype of this on another branch; it makes things appreciably faster.

src/perf_decode.ml
Svetlitski: I'm not entirely sure I understand what you're asking. Are you saying that interning `dso` immediately before using it on line 142 in `From_perf [%string "[unknown @ %{addr#Int64.Hex} (%{dso})]"]` is more efficient than interning it after usage? If so I agree, and it's a small change, but the way you've phrased this makes me think perhaps I'm missing something more significant.

Svetlitski: I made this change.

Svetlitski: Yes, changed.

src/event.ml
Svetlitski: Agreed; I changed it.

src/trace_segment.ml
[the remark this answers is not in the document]
Svetlitski: I added a comment where the field is initialized.

src/trace_segment.ml
```diff
@@ -0,0 +1,1080 @@
...
+  in
+  (* Then create the new frame for [dst]. *)
+  Nonempty_vec.push_back
+    t.callstacks
+    #{ time; leaf = Frame.create dst ~parent:(current_frame t); control_flow = Call }
+;;
+
+let handle_return (t : t) (time : Timestamp.t) ~(dst : Location.t) =
+  match (current_frame t).parent with
+  | Null ->
```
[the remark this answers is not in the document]
Svetlitski: When you've returned out of every known function, and are returning into something you never saw the `call` for. Your comment was made on an earlier version of the code; in the current implementation this is made more explicit:
```ocaml
let handle_return (t : t) (time : Timestamp.t) ~(dst : Location.t) =
  match current_physical_frame t with
  | #({ parent = Null; _ }, ~distance:_) -> return_to_unseen t time ~dst ~distance:0
  ...
```

src/trace_segment.ml
Svetlitski: I'm honestly not sure, I don't think I've seen this happen in my testing. And now that I think about it more, just grabbing a frame with a matching name from the current callstack if it does exist higher up feels wrong. I think I should probably just get rid of this case, and merge it with the `| #(Null, ~physical_distance:_, ..) ->` case as a last resort.

Svetlitski: Thoughts?

src/symbolizer.ml
Svetlitski: I modified the code to reflect this in the "Category" instead of modifying the function name. We don't even need to patch Perfetto, it shows the category of a slice when you hover over it.

src/new_trace_writer.ml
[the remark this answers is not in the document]
Svetlitski: Done. Makes `new_trace_writer.ml` considerably smaller.

src/trace_writer.ml
Svetlitski: In `New_trace_writer` the `finalize` and `end_of_trace` functions do different things; it's not streaming like `Trace_writer`, so this distinction matters for it. How about calling it `close` instead of `finalize`?

src/trace_segment.ml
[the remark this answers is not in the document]
Svetlitski: This is approximately what the code does now, but I found that it works in practice for the exception-heavy workloads I tested. FWIW this is without data loss though, it's the case where you start tracing at step 6., not where you've lost data.

src/trace_segment.ml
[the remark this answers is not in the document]
Svetlitski: I added a comment.

src/trace_segment.ml
Svetlitski: This PR is already gigantic, so I'll punt this to the follow-up with all of the other performance improvements.

src/trace_writer.ml
Xyene: Thanks for explaining; happy to leave it as-is.

src/trace_segment.ml
Svetlitski: Discussed offline, I added this back.

src/symbolizer.ml
```diff
@@ -0,0 +1,102 @@
...
+;;
+
+let symbolize t ~executable ~addr =
+  match executable with
+  | Null -> Null
+  | This executable ->
+    let addr = I64.of_int64 addr in
+    (* LLVM can't symbolize things in the Kernel; checking for this explicitly
+     avoids us polluting our cache with many [Null] responses. *)
+    if I64.O.(addr <= #0L)
```
Xyene, line 91: I think this should be `< 0`. `= 0` is technically fair game for userspace.

Xyene: (We already have `is_kernel_addr` defined somewhere, maybe worth lifting that somewhere common for clarity? Don't feel strongly about that, though.)

src/trace_segment.ml
```diff
@@ -0,0 +1,1650 @@
...
+      time
+      ~dso:new_location.dso
+      ~before:frame.instruction_pointer
+      ~after:new_location.instruction_pointer;
+    Frame.set_instruction_pointer frame new_location.instruction_pointer
+;;
+
+(** This is intended to mark code we hope is *unreachable*, not merely uncommon. Hopefully
+    over time we can strengthen its callsites to assertions. *)
+let[@cold] log_unexpected_case ~(here : [%call_pos]) sexp =
```
Xyene, line 580: When the original trace writer logged unexpected cases, it was often a "many times" or "none" situation, where an unexpected case would be hit hundreds/thousands of times in a trace, and rarely just once. This was spammy, so we rate-limited it:

https://github.com/janestreet/magic-trace/blob/62ce8e803fcc2e074ba2e004cd758024d9505024/src/trace_writer.ml#L159-L164

src/trace_segment.ml
```diff
@@ -0,0 +1,1650 @@
...
+  let #(src_frame, ~distance:_) = current_physical_frame t in
+  assert (not (Or_null.is_null src_frame.parent));
+  Frame.set_instruction_pointer src_frame src.instruction_pointer;
+  (* Then create the new frame for [dst]. *)
+  let dst_frame = Frame.create dst ~parent:(current_frame t) ~kind:Physical in
+  append_inlined_frames t time ~physical_frame:dst_frame ~physical_frame_is_new:true
+;;
+
+(** We are returning into something we did not see the call for. This can happen if
+    there's a series of calls like [fn1 -> fn2 -> fn3] and we started tracing during the
```
Xyene: Is the `-> fn3` meaningful in this example? I would have thought `fn1 -> fn2` + starting in `fn2` would be sufficient to hit this case.

src/trace_segment.ml
```diff
@@ -0,0 +1,1650 @@
...
+     instruction we are returning from, *not* whatever comes immediately after it. *)
+  let addr = Int64.O.(dst.instruction_pointer - 1L) in
+  let inlined_frames = symbolize_inlined_frames t ~dso:dst.dso ~addr in
+  let mutable inlined_leaf = Null in
+  for i = Slice.length inlined_frames - 1 downto 0 do
+    let%tydi inlined_frame_info = Slice.unsafe_get inlined_frames i in
+    let inlined_frame =
+      replace_root t (Symbolizer.Info.to_location inlined_frame_info) ~kind:Inlined
+    in
+    if Or_null.is_null inlined_leaf then inlined_leaf <- This inlined_frame
```
Xyene: Likely can simplify this control flow; it only ends up saving the first inlined frame.

src/trace_segment.ml
```diff
@@ -0,0 +1,1650 @@
...
+  let (physical_root : Frame.t) =
+    replace_root t { dst with instruction_pointer = addr } ~kind:Physical
+  in
+  Nonempty_vec.push_back
+    t.callstacks
+    #{ time
+     ; leaf = Or_null.value inlined_leaf ~default:physical_root
+     ; control_flow = Return { distance }
+     };
+  diff_inlined_frames t time ~dso:dst.dso ~before:addr ~after:dst.instruction_pointer;
```
Xyene, line 667: Nit: should `~dso` be called `~after_dso`?

src/trace_segment.ml
```diff
@@ -0,0 +1,1650 @@
...
+    [%test_eq: Interned_string.t or_null] current_physical_frame.location.dso dso);
+  diff_inlined_frames'
+    t
+    time
+    ~dso
+    ~before:(symbolize_inlined_frames t ~dso ~addr:before)
+    ~after:(symbolize_inlined_frames t ~dso ~addr:after)
+;;
+
+let append_inlined_frames t time ~(physical_frame : Frame.t) ~physical_frame_is_new =
```
Xyene: It's a little strange to me that a function called "`append_inlined_frames`" is doing more than just inline frame logic, specifically around inserting `physical_frame` into the callstack.

The reasoning for why we do this from the comment below makes sense to me, but I don't think is encapsulated in the name "`append_inlined_frames`".

src/trace_segment.ml
```diff
@@ -0,0 +1,1650 @@
...
+    if physical_frame_is_new || Slice.length inlined_frames > 0
+    then
+      (* It's important that a [physical_frame] and all of its inlined children are introduced
+         via a single [Callstack.t] for the sake of [smear_times] treating this correctly.
+         When control-flow proceeds from one address to another, where the destination address
+         corresponds to an inlined chain (e.g. [A -> B -> C]), you should treat all of the
+         functions in the chain as having been entered *simultaneously*.
+
+         Producing exactly one [Callstack.t] accomplishes this; if you instead produced N
+         [Callstack.t]s each of [Call { depth = 1}], time would get smeared between them. *)
```
Xyene:
```suggestion
         [Callstack.t]s each of [Call { depth = 1 }], time would get smeared between them. *)
```

src/trace_segment.ml
```diff
@@ -0,0 +1,1650 @@
...
+      let inlined_leaf_frame =
+        Frame.create
+          ~kind:Inlined
+          (Symbolizer.Info.to_location inlined_frame_info)
+          ~parent:!parent
+      in
+      parent := inlined_leaf_frame);
+    if physical_frame_is_new || Slice.length inlined_frames > 0
+    then
+      (* It's important that a [physical_frame] and all of its inlined children are introduced
```
Xyene: Worth explaining that not doing this won't _break_ anything, but it'll make the trace look worse.

src/symbolizer.ml
Svetlitski: You're never going to symbolize code at `NULL`.

Svetlitski: I'll update the comment.

src/trace_segment.ml
Svetlitski: I agree your example works too. I suppose I used 3 functions in hopes of making it clearer that the return into `fn1` needs to be handled specially compared to the return into `fn2`.

src/trace_segment.ml
Svetlitski: I updated the code to reflect this.

src/trace_segment.ml
Svetlitski: `before` and `after` need to be addresses within the same physical symbol in order for the call to be well-formed, so they have to be in the same `dso` too, hence just "`dso`".

src/trace_segment.ml
```diff
@@ -0,0 +1,1663 @@
...
+    execution of [fn2], then we see a return into [fn1]. *)
+let return_to_unseen (t : t) (time : Timestamp.t) ~(dst : Location.t) ~(distance : int) =
+  (* We symbolize at one byte before the return address because the callstack we want to
+     see reflected as roots of the whole trace corresponds to the code as of the [call]
+     instruction we are returning from, *not* whatever comes immediately after it. *)
+  let addr = Int64.O.(dst.instruction_pointer - 1L) in
+  let inlined_frames = symbolize_inlined_frames t ~dso:dst.dso ~addr in
+  let mutable inlined_leaf = Null in
+  for i = Slice.length inlined_frames - 1 downto 0 do
+    let%tydi inlined_frame_info = Slice.unsafe_get inlined_frames i in
```
Xyene: Is the `let%tydi` doing anything here, given we're not destructuring?

Svetlitski: Thanks for catching this, this appears to be vestigial. I removed it.

src/trace_segment.ml
Svetlitski: I updated the comment; I took a slightly more aggressive stance here that doing the opposite would arguably be incorrect.

src/trace_segment.ml
Svetlitski: Modified.

src/trace_segment.ml
Svetlitski: I could keep the existing `append_inlined_frames` function unchanged as the underlying implementation, but immediately shadow it with two functions:
```ocaml
let insert_frame_and_append_inlined_frames t time ~physical_frame =
  append_inlined_frames_impl t time ~physical_frame ~physical_frame_is_new:true
;;

let append_inlined_frames t time ~physical_frame = 
  append_inlined_frames_impl t time ~physical_frame ~physical_frame_is_new:false
;;
```

Or something roughly like that if you prefer.

src/trace_segment.ml
Svetlitski: While this code is still so new, I want to know if we're hitting this once or a thousand times though when we're addressing any initial user reports.

src/trace_segment.ml
```diff
@@ -0,0 +1,1080 @@
...
+        (#{ last_callstack with control_flow = Return { distance = 0 } } : Callstack.t)
+  ; exception_handlers = Vec.copy existing.exception_handlers
+  ; in_filtered_region
+  }
+;;
+
+let in_filtered_region t = t.in_filtered_region
+let[@inline always] current_frame t = (Nonempty_vec.last t.callstacks).#leaf
+
+let replace_root t location =
```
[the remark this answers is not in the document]
Svetlitski: It mutates the contents of `t.root` to turn it into a standard `Frame.t` at `location`, then replaces `t.root` with a new `Frame.Sentinel.t` which sits above it.

That being said, I do agree "replace" as a verb doesn't correctly set the expectation that calling this function 3 times increases the depth of the stack by 3.

How about `emplace_root`, taking inspiration from C++ where "emplace" has the connotation of "constructing a new element in-place", e.g. `std::vector::emplace_back`? I'll make this change in the meantime, but lmk if you're still unhappy with that.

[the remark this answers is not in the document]
Xyene: Sure, that sounds good to me.

src/trace_segment.ml
Svetlitski: Removed `fn3` from the comment.

src/trace_segment.ml
Svetlitski: Resolved offline, renamed accordingly.

Xyene (approved):
LGTM, thanks a lot for working on this!

PR #356, opened 2026-04-15
Fix race condition on frequently hit breakpoint

Svetlitski:
Specifying a breakpoint `-trigger SYMBOL` on a `SYMBOL` that is very frequently executed would oftentimes result in magic-trace waiting indefinitely, never taking a snapshot. The root-cause was that by default we configure breakpoints as single-hit, but if the first hit occurs between the time where magic-trace creates the breakpoint, and the time where it registers the breakpoint's file descriptor with `epoll`, magic-trace never sees the notification.

This is fixed by only enabling the breakpoint after registering its file descriptor with `epoll`.

Disclaimer: The initial commit was written entirely by Claude Opus 4.6, but I performed manual cleanup.

### DIFF
--- a/src/breakpoint.ml
+++ b/src/breakpoint.ml
@@ -5,7 +5,6 @@ type t
 external create
   :  pid:Pid.t
   -> addr:int64
-  -> single_hit:bool
   -> (t, int) result
   = "magic_breakpoint_create_stub"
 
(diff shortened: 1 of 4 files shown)


cgaebel (approved):
lgtm

PR #363, opened 2026-05-01
Fix stack switching on OpenCilk's runtime

kennyzzhang:
OpenCilk's work stealing scheduler [cheetah](https://github.com/OpenCilk/cheetah) uses longjmp to switch between user and runtime code. This causes the stackframes to stairstep downwards, as expected from the FAQ.

image

So, I've "added some explicit code in magic-trace for [OpenCilk]".

In particular, upon jumping to user code, the runtime's stack is put into an inactive stack, which is then restored when jumping back to the user code.

The main two functions in cheetah that call longjmp are `longjmp_to_runtime` and `sysdep_longjmp_to_sf`.
`sysdep_longjmp_to_sf` is called from two different places, `longjmp_to_user_code` or `__cilkrts_sync`.
This means an extra frame needs to be popped when `sysdep_longjmp_to_sf` is called, and we should only put the runtime stack into an inactive stack when we see `longjmp_to_user_code`.

This is my first time writing OCaml, and honestly I was fumbling around copying and adapting other parts of the codebase to coerce things to work. But it seems like things work well enough that it's now much easier to tell if Cilk workers are failing to find work.

image

### DIFF
--- a/src/trace_writer.ml
+++ b/src/trace_writer.ml
@@ -604,18 +604,18 @@ let end_of_thread t (thread_info : _ Thread_info.t) ~time ~is_kernel_address : u
   Thread_info.set_callstack thread_info ~is_kernel_address ~time
 ;;
 
-(* OpenCilk's runtime cheetah (https://github.com/OpenCilk/cheetah) uses
-   longjmp to switch between user and runtime stacks.
+(* OpenCilk's runtime cheetah (https://github.com/OpenCilk/cheetah) uses longjmp to switch
+   between user and runtime stacks.
 
-   To deal with this stack switching, when jumping into the runtime, we need
-   to clear the user's stack frames. The inactive_callstacks mechanism
-   is perfect for keeping track of this stack data. *)
+   To deal with this stack switching, when jumping into the runtime, we need to clear the
+   user's stack frames. The inactive_callstacks mechanism is perfect for keeping track of
+   this stack data. *)
 module OpenCilk_hacks : sig
   val call_handle_stack_switch
-    : 'a inner
+    :  'a inner
     -> 'a Thread_info.t
     -> time:Mapped_time.t
-    -> location: Event.Location.t
+    -> location:Event.Location.t
     -> unit
 end = struct
   let ret = ret_without_checking_for_go_hacks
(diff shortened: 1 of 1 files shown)


src/trace_writer.ml
```diff
@@ -610,6 +604,66 @@ let end_of_thread t (thread_info : _ Thread_info.t) ~time ~is_kernel_address : u
...
+    :  'a inner
+    -> 'a Thread_info.t
+    -> time:Mapped_time.t
+    -> location:Event.Location.t
+    -> unit
+end = struct
+  let ret = ret_without_checking_for_go_hacks
+
+  let switch_to_user_code t (thread_info : _ Thread_info.t) ~time =
+    (* Pop the sysdep_longjmp_to_sf frame *)
```
Xyene, line 624:
```suggestion
    (* Pop the [sysdep_longjmp_to_sf] frame. *)
```

src/trace_writer.ml
```diff
@@ -610,6 +604,66 @@ let end_of_thread t (thread_info : _ Thread_info.t) ~time ~is_kernel_address : u
   Thread_info.set_callstack thread_info ~is_kernel_address ~time
 ;;
 
+(* OpenCilk's runtime cheetah (https://github.com/OpenCilk/cheetah) uses longjmp to switch
+   between user and runtime stacks.
+
+   To deal with this stack switching, when jumping into the runtime, we need to clear the
+   user's stack frames. The inactive_callstacks mechanism is perfect for keeping track of
```
Xyene, line 611:
```suggestion
   user's stack frames. The [inactive_callstacks] mechanism is perfect for keeping track of
```

src/trace_writer.ml
```diff
@@ -610,6 +604,66 @@ let end_of_thread t (thread_info : _ Thread_info.t) ~time ~is_kernel_address : u
...
+    -> time:Mapped_time.t
+    -> location:Event.Location.t
+    -> unit
+end = struct
+  let ret = ret_without_checking_for_go_hacks
+
+  let switch_to_user_code t (thread_info : _ Thread_info.t) ~time =
+    (* Pop the sysdep_longjmp_to_sf frame *)
+    ret t thread_info ~time;
+    (* The next stack frame is either __cilkrts_sync or longjmp_to_user_code. In either
```
Xyene, line 626:
```suggestion
    (* The next stack frame is either [__cilkrts_sync] or [longjmp_to_user_code]. In either
```

src/trace_writer.ml
```diff
@@ -610,6 +604,66 @@ let end_of_thread t (thread_info : _ Thread_info.t) ~time ~is_kernel_address : u
...
+    match Callstack.top thread_info.callstack with
+    | Some { symbol = From_perf symbol; _ } ->
+      (match symbol with
+       | "longjmp_to_user_code(__cilkrts_worker*, Closure*)" ->
+         ret t thread_info ~time;
+         Stack.push thread_info.inactive_callstacks thread_info.callstack;
+         thread_info.callstack <- Callstack.create ~create_time:time
+       | "__cilkrts_sync" -> ret t thread_info ~time
+       | _ -> Printf.printf "OpenCilk_hacks: Unexpected symbol %s\n" symbol)
+    | _ -> Printf.printf "OpenCilk_hacks: Unexpected symbol [unknown]"
```
Xyene, line 637: Nit: these are not just `[unknown]` symbols. Writing the match exhaustively would make that a little more clear.

src/trace_writer.ml
```diff
@@ -610,6 +604,66 @@ let end_of_thread t (thread_info : _ Thread_info.t) ~time ~is_kernel_address : u
...
+       case, we want to pop this frame as well. *)
+    match Callstack.top thread_info.callstack with
+    | Some { symbol = From_perf symbol; _ } ->
+      (match symbol with
+       | "longjmp_to_user_code(__cilkrts_worker*, Closure*)" ->
+         ret t thread_info ~time;
+         Stack.push thread_info.inactive_callstacks thread_info.callstack;
+         thread_info.callstack <- Callstack.create ~create_time:time
+       | "__cilkrts_sync" -> ret t thread_info ~time
+       | _ -> Printf.printf "OpenCilk_hacks: Unexpected symbol %s\n" symbol)
```
Xyene, line 636: Nit: let's use `eprintf` here, and use the `%!` syntax to force a flush. For example, see

https://github.com/kennyzzhang/magic-trace/blob/c31a2f752fbdf41f786b87b3c0fd5ff4869b05eb/src/trace_writer.ml#L1027-L1034

src/trace_writer.ml
```diff
@@ -610,6 +604,66 @@ let end_of_thread t (thread_info : _ Thread_info.t) ~time ~is_kernel_address : u
   Thread_info.set_callstack thread_info ~is_kernel_address ~time
 ;;
 
+(* OpenCilk's runtime cheetah (https://github.com/OpenCilk/cheetah) uses longjmp to switch
```
Xyene, line 607: Nit: let's move all of this code up, so that `let call ... = ...` is close in the file with `let ret ... = ...`.

Xyene (commented):
Thanks, looks mostly good! A couple of nits/suggestions.

PR #364, opened 2026-05-04
Add C++ symbol demangling via __cxa_demangle

AneesPatel:
C++ symbols (`_Z...`) currently render raw in flamegraphs. Anything templated is unreadable.

This PR wires `__cxa_demangle` into `Symbol_resolver.resolve`, which is the single path every trace backend goes through, so readable names show up everywhere at once.

A few notes on the implementation:

* The C stub passes `NULL`/`NULL` to `__cxa_demangle` so it heap-allocates the output itself. A stack buffer isn't safe here: the demangler may `realloc` its argument when it's too small, and `realloc`'ing stack memory is UB.
* `Cxx_demangle.demangle` checks for the `_Z` prefix in OCaml before crossing the FFI boundary, so traces dominated by non-C++ symbols don't pay the FFI cost.
* Linked against `-lsupc++` rather than `-lstdc++`. The stub doesn't use any C++ stdlib features, and `libsupc++` ships `__cxa_demangle` on its own without pulling in the `GLIBCXX_*` versioning surface. Keeps static release builds simpler.
* Fallback chain: C++ demangle, then `Demangle_ocaml_symbols.demangle`, then the raw string. Frames are never dropped.

One thing this doesn't cover: Rust v0 (`_R...`, default since rustc 1.77). `__cxa_demangle` handles legacy Rust by accident because of the shared Itanium prefix, but anything from a recent rustc won't demangle. Probably worth a follow-up PR.


kiwigitops (commented):
Nice — the `_Z` prefix short-circuit in `Cxx_demangle.demangle` is a good touch for keeping the FFI off the hot path for non-C++ symbols.

One question on the integration in `elf.ml` (`Symbol_resolver.resolve`): demangling now happens on every call to `resolve`, and since each call does a fresh `malloc` + libc++abi parse inside `__cxa_demangle`, the same mangled name (e.g. a deeply-templated `std::variant<...>` instantiation that appears in many frames) will be re-demangled each time it's encountered. Did you measure the cost on a representative C++ trace, or consider memoizing by mangled name inside `Symbol_resolver` (or one level up)? Curious whether it's negligible in practice or worth a small hashtable — happy either way, just wanted to flag it since the PR description notes this is the single path all backends share.

Also a small naming nit, take it or leave it: the test name `_Z-prefixed garbage fails gracefully` in `cxx_demangle.ml` is great, but it might be worth adding one positive integration-level test through `Symbol_resolver.resolve` so the wiring is regression-protected, not just the FFI in isolation.

Svetlitski:
We'll get this functionality as a side-effect of #353 because we switch to using LLVM for symbolization, so I'm closing this PR in anticipation of landing that.

PR #366, opened 2026-05-06
Remove the perf ctlfd timeout

lnbrks:
The operations perf is doing behind the scenes are effectively unbounded depending on the parameters we pass it. We encountered an issue with -multi-thread causing the snapshot itself to take too long.

If there's a real need for a timeout in the future, we can make it configurable.


Xyene (approved):
LGTM, thanks!

PR #367, opened 2026-05-06
OCaml stack switching

TheNumbat:
(WIP, needs cleanup)

Handles OCaml 5 effect primitives by shifting/restoring the location and exception handler stack when entering and exiting fibers.
When available, interprets ptwrite events from perform and resume to generate a flow arrow from the suspention to the resumption.

Based on #328 + #353. Also see https://github.com/oxcaml/oxcaml/pull/6032.


TheNumbat:
Moved to #376

PR #371, opened 2026-05-29
Update OCaml symbol name resolution

Svetlitski:
Now that [PR 5976](https://github.com/oxcaml/oxcaml/pull/5976) has made it into the latest OxCaml compiler release, we need to change the way we resolve OCaml symbol names. We continue to use the compiler-emitted "pretty" name as before; the updates here just handle the fact that has been moved from `DW_AT_linkage_name` to `DW_AT_name`.

### DIFF
--- a/src/dune
+++ b/src/dune
@@ -15,7 +15,7 @@
  (action
   (progn
    (bash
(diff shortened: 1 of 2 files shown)


Xyene (approved):
LGTM, thanks!

PR #372, opened 2026-05-30
Parse uppercase flags in perf CBR events

### LINKED ISSUE #324: "exception raised while parsing perf output" with "cbr: 36 freq: 3629 MHz (450%)"
With Ubuntu 24.04, magic-trace 1.2.4 fails to parse the perf data it records for a simple test application when `-trace-include-kernel` is specified:

```
root@eddie-beman-server:~# /home/eddie/magic-trace --version
v1.2.4
root@eddie-beman-server:~# perf --version
perf version 6.8.12
root@eddie-beman-server:~# head -n 1 /etc/os-release
PRETTY_NAME="Ubuntu 24.04.1 LTS"
root@eddie-beman-server:~# head -n 5 /proc/cpuinfo
processor       : 0
vendor_id       : GenuineIntel
cpu family      : 6
model           : 190
model name      : Intel(R) N150
root@eddie-beman-server:~# echo 'int main() { for (int i = 0; i < 1000000000; ++i) {} }' > foo.c && gcc -O0 ./foo.c -o foo
root@eddie-beman-server:~# MAGIC_TRACE_DEBUG=1 /home/eddie/magic-trace run -z-print-perf-commands -trace-include-kernel ./foo
perf record -o /tmp/magic_trace.tmp.27fbc1/perf.data --timestamp --event=intel_pt/cyc=1,cyc_thresh=1,mtc_period=0/uk --per-thread -t 377433 --snapshot=e --kcore
[ Snapshot taken. ]
[ perf record: Captured and wrote 4.015 MB /tmp/magic_trace.tmp.27fbc1/perf.data ]
[ Finished recording. ]
perf script -i /tmp/magic_trace.tmp.27fbc1/perf.data --ns --itrace=bep -F pid,tid,time,flags,ip,addr,sym,symoff,synth,dso,event,period --dlfilter /tmp/magic_trace.tmp.27fbc1/perf_dlfilter.so
(monitor.ml.Error
 ("BUG: exception raised while parsing perf output. Please report this to https://github.com/janestreet/magic-trace/issues/"
  (exn Not_found)
  (perf_output
   (" 377433/377433  1693789.232275122:          1                                         cbr:   D                      cbr: 36 freq: 3629 MHz (450%)                   0     6143f240a145 main+0x1c (/root/foo)")))
  "Called from Async_kernel__Pipe.map.(fun) in file \"src/pipe.ml\", line 1066, characters 65-70"
  "Called from Async_kernel__Job_queue.run_jobs in file \"src/job_queue.ml\", line 180, characters 6-47"
```

jeet-dekivadia:
## Summary
- accept uppercase letters in the ignored annotation prefix of perf CBR events
- add a regression using the Ubuntu 24.04 `perf script` line from the issue

## Why
Newer perf output can include an uppercase `D` marker before the CBR payload. The parser only accepted lowercase annotation text there, so `Re.exec` raised `Not_found` before the frequency event could be decoded.

## Validation
- `git diff --check`
- exercised the widened CBR pattern against the reported line and confirmed it extracts `3629` MHz
- added expect coverage for the reported parser path

Fixes #324

### DIFF
--- a/src/perf_decode.ml
+++ b/src/perf_decode.ml
@@ -26,7 +26,8 @@ let perf_branches_event_re =
 ;;
 
 let perf_cbr_event_re =
-  Re.Perl.re {|^ *([a-z )(]*)? +cbr: +([0-9]+ +freq: +([0-9]+) MHz)?(.*)$|} |> Re.compile
+  Re.Perl.re {|^ *([a-zA-Z )(]*)? +cbr: +([0-9]+ +freq: +([0-9]+) MHz)?(.*)$|}
+  |> Re.compile
 ;;
 
 let trace_error_re =
@@ -646,6 +647,18 @@ module%test _ = struct
         ((Ok
           ((thread ((pid (21302)) (tid (21302)))) (time 22h51m58.700445693s)
            (data (Power (freq 4500)))))) |}]
+  ;;
+
+  let%expect_test "cbr event with uppercase flag" =
+    check
+      "377433/377433 1693789.232275122:          1                                  \
+       cbr:   D                      cbr: 36 freq: 3629 MHz (450%)                   \
+       0     6143f240a145 main+0x1c (/root/foo)";
+    [%expect
+      {|
+        ((Ok
+          ((thread ((pid (377433)) (tid (377433)))) (time 19d14h29m49.232275122s)
+           (data (Power (freq 3629)))))) |}]
   ;;
 
   (* Expected [None] because we ignore these events currently. *)


jeet-dekivadia:
I rechecked this today while it is waiting for review. The parser change is limited to accepting uppercase annotation flags in perf CBR event lines, matching the shape emitted by newer perf versions such as Ubuntu 24.04's `D` flag output.

The extraction path for the CBR value is otherwise unchanged; the regression keeps the reported uppercase-flag line decoding to the expected MHz value. DCO is green, and the PR has no red checks.

PR #374, opened 2026-05-30
Open attached executables through procfs

### LINKED ISSUE #307: Failure to read symbols inside docker-container due to broken /proc/pid/exe link
Hi. I encountered this problem while trying to trace binary inside docker-container.
The problem manifests as rather misleading
```
Cannot select a snapshot symbol because magic-trace can't find that executable's symbol table. Was it built without debug info, or with debug info magic-trace doesn't understand?
```
message which doesn't indicate failure to open the executable but strace helped here to find the root cause as running the binary outside of container worked well.

The reason is that `/proc/<pid>/exe` contains a path used to launch the process _inside_ the container which means this path refers to directories _inside_ the container. This symlink is not valid outside of the container

I believe the problem can be fixed by skipping `Core_unix.readlink` here https://github.com/janestreet/magic-trace/blob/7ac635e41438248c3a3348c039dae00854f38342/src/trace.ml#L722 and just directly read the contents of the `/proc/<pid>/exe`.

jeet-dekivadia:
## Summary
- keep `/proc/<pid>/exe` intact while loading an attached process executable
- use the procfs symlink target only when matching the executable name reported by `/proc/<pid>/maps`
- preserve the existing `realpath` behavior for ordinary executable paths

Fixes #307.

## Why both paths are needed
For a process in another mount namespace, `/proc/<pid>/exe` remains openable from the host even when its symlink target, such as `/app/server`, is not present in the host filesystem. Concretely, `readlink("/proc/123/exe")` can return `/app/server` while `open("/app/server")` fails on the host and `open("/proc/123/exe")` succeeds. The maps file still reports `/app/server`, so stop-filter address calculation needs the `readlink` result for its pathname comparison while ELF loading needs the procfs path itself.

## Validation
- `git diff --check`
- reviewed both executable consumers: ELF loading now receives the procfs path for attach mode, and map-name comparison resolves only procfs executable symlinks with `readlink`

### DIFF
--- a/src/elf.ml
+++ b/src/elf.ml
@@ -302,7 +302,14 @@ let all_file_selections t symbol =
 ;;
 
 let selection_stop_info t pid selection =
-  let filename = Filename_unix.realpath t.filename in
+  (* Keep procfs executable symlinks openable across mount namespaces, but compare
+     against the target path exposed in [/proc/$pid/maps]. *)
+  let filename =
+    if String.is_prefix t.filename ~prefix:"/proc/"
+       && String.is_suffix t.filename ~suffix:"/exe"
+    then Core_unix.readlink t.filename
+    else Filename_unix.realpath t.filename
+  in
   let compute_addr addr =
     if t.statically_mappable
     then addr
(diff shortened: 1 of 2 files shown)


jeet-dekivadia:
I rechecked this today while it is waiting for review. The patch stays on the attach path: when magic-trace is reading symbols for an already-running target, it opens the executable through procfs so the path is resolved in the target process context instead of the tracer's current mount namespace.

The startup/non-attach path still uses the existing executable resolution. DCO is green, and the PR has no red checks.

PR #375, opened 2026-05-30
Stop attach sessions when perf exits

### LINKED ISSUE #312: magic-trace does not stop running on process SIGKILL
When running magic-trace to trace a process, when SIGKILL is sent to the traced process, the magic-tracer does not stop running, while `perf record` does stop. I expected the magic tracer to stop running or to take a snapshot immediately upon the process finishing. To reproduce the issue:
1. Run `sleep 5000` and get its PID.
2. Run `magic-trace attach -p <pid> -o magic-trace.fxt`
3. Run `kill -9 <pid>` to see that `perf record` stops running while `magic-trace` does not.

Running version 6.7.0 of perf userspace tools.

jeet-dekivadia:
## Summary

- cache the perf recorder completion deferred so monitoring and final cleanup share a single `waitpid`
- wake attach sessions when their recorder exits after the tracee terminates
- tolerate `ESRCH` when snapshot or shutdown signals race an already-exited recorder

## Why

In attach mode, killing the traced process makes `perf record` exit, but magic-trace only waits for Ctrl-C or a breakpoint snapshot. That leaves the command hanging after the tracee is gone. Sharing the recorder completion result also keeps shutdown from attempting a second reap.

Fixes #312

## Validation

- `git diff --check`
- reviewed the attach wakeup and cached `waitpid` flow statically
- verified `Perf_tool_backend.Recording` is the only `Backend_intf.S.Recording` implementation and that monitoring plus final cleanup share the same cached `waitpid` deferred

### DIFF
--- a/src/backend_intf.ml
+++ b/src/backend_intf.ml
@@ -32,6 +32,7 @@ module type S = sig
       -> (t * Data.t) Deferred.Or_error.t
 
     val maybe_take_snapshot : t -> source:[ `ctrl_c | `function_call ] -> unit
+    val finished : t -> unit Deferred.t
     val finish_recording : t -> unit Deferred.Or_error.t
   end
 
--- a/src/perf_tool_backend.ml
+++ b/src/perf_tool_backend.ml
@@ -119,8 +119,8 @@ module Recording = struct
        sends SIGINT to magic-trace.
 
        In run mode, it is guaranteed via waitpid that the tracee exits before backend
-       snapshotting logic runs. In attach mode, magic-trace doesn't get any notification when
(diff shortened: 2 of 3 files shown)


jeet-dekivadia:
Small follow-up for review context. This remains scoped to attach-session shutdown: `Perf_tool_backend.Recording` is the only recording implementation, and the patch caches its completion deferred so monitoring and cleanup observe the same single `waitpid` result. That lets a terminated tracee wake attach mode without changing Ctrl-C or breakpoint behavior, while the `ESRCH` handling only covers signal races after recorder exit.

PR #376, opened 2026-06-15
OxCaml Stack Switching

TheNumbat:
Adds special handling for OCaml 5 effect primitives. 

In OxCaml, if the `ptwrite` instruction is supported, the runtime will emit events that record the fiber ID upon perform and resume. Magic trace can use this ID to save and restore the fiber's callstack and exception handling state. We now also generate flow events connecting the perform and resume pair. Support for `ptwrite` events comes from #328. 

If `ptwrite` events are not present, we still try to track state for the current fiber, which means programs that perform and resume with a stack discipline will work correctly. Otherwise, we still use the most recent state, so callstacks may be incorrect, but the trace still mostly works (e.g. the `migrate` test hits unknown return warnings, but still produces something coherent). 

Adds several sample programs that exercise effects and exceptions. They all give the expected structure with `ptwrite` events and something coherent without.

There are currently a bunch of hardcoded offsets to detect the effect operations - I'll add these to the compiler output (alongside the exn handler info) and read them in another pr.


glittershark:
@TheNumbat this will need to be updated for https://github.com/oxcaml/oxcaml/pull/6440

src/event.ml
```diff
@@ -88,6 +88,10 @@ module Ok = struct
           ; count : int
           ; name : Collection_mode.Event.Name.t
           }
+      | Ptwrite of
+          { location : Location.t
+          ; data : string
```
Svetlitski: Why is `data` of type `string`? The `ptwrite` instruction can only hold a 32-bit/64-bit payload, and judging by `perf_ptwrite_event_re` in `src/perf_decode.ml` it appears `perf script` renders it as a number.

src/trace_segment.ml
```diff
@@ -354,6 +358,14 @@ module Callstack = struct
      }
 end
 
+module Fiber = struct
```
Svetlitski, line 366: This could really use a doc-comment explaining things like:
- What does a `Fiber.t` represent?
- Why do we need special handling for them?
- etc.

src/trace_segment.ml
```diff
@@ -373,6 +385,9 @@ type t =
   ; symbolizer : Symbolizer.t
   ; ocaml_exception_info : Ocaml_exception_info.t or_null
   ; exception_handlers : Frame.t Vec.t
+  ; effect_handlers : (Frame.t * exn_depth:int) Vec.t
```
Svetlitski: Doc-comments explaining these fields would be great.

src/trace_segment.ml
```diff
@@ -850,47 +854,173 @@ let handle_ocaml_exception (t : t) (time : Timestamp.t) ~(dst : Location.t) =
...
+        (* This is subtle; Yes the frame is new, but we *don't* want to reflect that in a
+              [Call], because discovered roots are handled separately. *)
+      ~and_insert_physical_frame_too:false
+;;
+
+let handle_ocaml_exception (t : t) (time : Timestamp.t) ~(dst : Location.t) =
+  match Vec.last t.exception_handlers with
+  | This dst_frame ->
+    Vec.pop_back_unit_exn t.exception_handlers;
+    if not (Symbol.equal dst_frame.location.symbol dst.symbol)
```
Svetlitski, line 912: Why did you weaken this from an `assert` to a log? I'd prefer to keep this strict, getting this wrong is generally evidence of a correctness issue.

src/trace_segment.ml
```diff
@@ -850,47 +854,173 @@ let handle_ocaml_exception (t : t) (time : Timestamp.t) ~(dst : Location.t) =
...
+      t
+      time
+      ~new_location:dst
+      ~frame
+      ~distance
+      ~leaf_of_inlined_stack
+  | #(maybe_frame, ~distance, ..) ->
+    (* We are probably raising into an exception handler much further up the stack that we never saw the entrance into. *)
+    let distance =
+      (* - Add 1 to the distance for the [_phantom_frame] we are injecting.
```
Svetlitski, line 876: Could you please restore the comment indentation to what it was previously? It looks like is an issue in most of the existing comments you moved around.

src/trace_segment.ml
```diff
@@ -14,7 +14,8 @@ module Frame : sig
       | Inlined (** An inlined function call, known to us solely via debug-info. *)
   end
 
-  (* The [location], [parent], and [kind] fields are actually **immutable** except for on [Sentinel.t] instances. *)
+  (* The [location], [parent], and [kind] fields are actually **immutable** except for on [Sentinel.t] instances.
```
Svetlitski: `parent` being immutable save for on `Sentinel.t`s (when `parent` is `Null`) was vital for correctness, because the invariant that all `Frame.t` chains in a trace-segment end with the same `Sentinel.t` is what grants you the ability to do an O(1) update to place a new parent frame at the top of stack and know that it applies to all the frames in the trace-segment. You don't get this guarantee with non-sentinel frames. Are you sure this is not broken?

src/trace_writer_intf.ml
```diff
@@ -45,4 +45,10 @@ module type S_trace = sig
     -> name:string
     -> time:Time_ns.Span.t
     -> unit
+
+  type flow
```
Svetlitski: Minor nit: `module Flow : sig type t end` would be preferable.

Svetlitski (changes requested):

src/event.ml
TheNumbat: I don't see any particular reason, fixed

src/trace_segment.ml
TheNumbat: I think this was just for debugging, switched to failwith. However, I think both this failwith and the `exception handler not found in the current callstack` case can be hit if we start tracing between the handler and the exception, right? (Same applies to switching to a previously-unknown fiber and then raising into a handler installed on that fiber.)

src/trace_segment.ml
TheNumbat: oops, fixed

src/trace_segment.ml
TheNumbat: I believe it's still true that all `Frame.t` chains in a trace segment end with the same sentinel. Suspending a callstack by packaging it into a `Fiber.t` removes the callstack from the trace segment, and resuming it splices it back into the correct path to the current sentinel. Indeed `reparent` can easily violate this invariant, so I at least renamed it and expanded the comment

Svetlitski: The sentinel being the common ancestor of `Frame.t` chains is the *reason* it's the only `Frame.t` whose `parent` can be modified safely; for any other `Frame.t`, we have no guarantee that this reparenting will apply to every `Frame.t` chain in the `Trace_segment`. Here's a case where this breaks things:
```
1  Call  perform0            -> caml_perform
2  Jump  caml_perform+0xb5   -> caml_apply3      # perform hook, effect_handlers empty
3  Jump  caml_apply3         -> effc
4  Call  effc                -> continue
5  Call  continue            -> caml_resume
6  Jump  caml_resume+0xda    -> fun              # resume hook

Output of write_trace:

  0ns ENTER caml_apply3
  0ns ENTER [zero or more unknowable frames]
  1ns ENTER perform0
  1ns ENTER caml_perform
  2ns EXIT  caml_perform
write_trace RAISED: "got unexpected result" ((expected (From_perf perform0)) (got (From_perf continue)))
```

If you modify the parent of anything besides the sentinel, when writing out the trace the series of frames we enter via `Call`s is no longer guaranteed to match the series of frames we exit via `Returns`.

PR #379, opened 2026-06-24
Support running commands while attached

### LINKED ISSUE #230: Support `magic-trace attach -p ... [cmd]` syntax
This would allow doing e.g.

```
$ magic-trace attach -p $(pidof foo) -trigger bar sleep 5
```

like `perf`.

yinli-systems:
Fixes #230.

## Summary

Adds an optional command argument to `magic-trace attach`, so users can attach to an existing process, run a workload, and have magic-trace detach automatically when that workload exits:

```sh
magic-trace attach -p $PID -- ./load-generator arg1 arg2
```

The existing Ctrl-C-driven attach flow is unchanged. The new path reuses the existing attach/record/decode pipeline, forwards the helper command's stdio, detaches when the helper exits, and still detaches promptly if magic-trace receives Ctrl-C first.

## Validation

- `git diff --check`
- `make` attempted, but this local machine does not have `dune` installed (`make: dune: No such file or directory`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ You can point magic-trace at a function such that when your application calls it
 
 1. [Here](https://raw.githubusercontent.com/janestreet/magic-trace/master/demo/demo.c)'s a sample C program to try out. It's a slightly modified version of the example in `man 3 dlopen`. Download that, build it with `gcc demo.c -ldl -o demo`, then leave it running `./demo`. We're going to use that program to learn how `dlopen` works.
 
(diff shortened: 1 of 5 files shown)


yinli-systems:
@Svetlitski, could you review this when convenient? The main design question is the attach/workload lifecycle: command exit detaches while preserving failures; recording-first completion terminates and reaps the command process group; simultaneous completion keeps the command result. The no-command path is unchanged. The inline race/process tests, `@fmt`, x86_64 static build, and full static-profile test suite all pass; the latest commit also fixes the new internal library for the repository's static profile with `no_dynlink`.

PR #380, opened 2026-07-04
Detect perf capabilities for kernel tracing

### LINKED ISSUE #94: Better capability checking for if we can trace the kernel
Ref https://perf.wiki.kernel.org/index.php/Perf_tools_support_for_Intel%C2%AE_Processor_Trace#Adding_capabilities_to_perf

Right now we just check if the user is root, which covers most cases but isn't precise.

Abhishek21g:
## Summary

- detect effective `cap_perfmon` / `cap_sys_admin` file capabilities on the configured `perf` executable
- keep root as the fast path for kernel tracing permissions
- fall back to the previous root-only behavior when `getcap` is unavailable or cannot inspect `perf`
- update the kernel tracing permission error to mention file capabilities
- update the README wording around kernel tracing permissions
- add expect tests for `getcap` output parsing

Fixes #94.

## Notes

This keeps the existing capability detection structure and only broadens the kernel-tracing permission check. If `getcap` is unavailable or cannot inspect the configured `perf`, magic-trace keeps the previous root-only behavior.

## Validation

- `git diff --check`
- DCO check is passing on GitHub
- Linux validation on fork branch with this patch plus a temporary workflow file: https://github.com/Abhishek21g/magic-trace/actions/runs/28693698343
  - `dune build @fmt`
  - `make PROFILE=static`
  - `dune runtest --profile=static`

Optional non-root capability smoke test on a disposable Linux machine:

```sh
sudo setcap cap_perfmon,cap_sys_ptrace,cap_syslog=ep "$(command -v perf)"
getcap "$(command -v perf)"

MAGIC_TRACE_PERF_PATH="$(command -v perf)" \
  ./_build/default/bin/magic_trace_bin.exe run -trace-include-kernel true
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ You can point magic-trace at a function such that when your application calls it
 
 Congratulations, you just magically traced your first program!
 
-In contrast to traditional `perf` workflows, magic-trace excels at hypothesis generation. For example, you might notice that taking 6us to run `cos` is a really long time! If you zoom in even more, you'll see that there's actually five pink "\[untraced\]" cells in there. If you re-run magic-trace with root and pass it `-trace-include-kernel`, you'll see stacktraces for those. They're page fault handlers! The demo program actually calls `cos` twice. If you zoom in even more near the end of the 6us `cos` call, you'll see that the second call takes *far* less time and does not page fault.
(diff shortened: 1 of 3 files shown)


Abhishek21g:
I noticed the older related PR #254 after opening this. Since that PR appears stale and #94 is still open, I put together a smaller version that keeps the existing capability detection structure, falls back to the current root-only behavior when getcap is unavailable, and adds focused expect tests for the parser. Happy to adjust direction if maintainers prefer reviving #254.

PR #386, opened 2026-08-14
Add Intel PT AUX trace range support

namtran1812:
## Summary

Add record-time Intel PT range gating using perf AUX actions.

This introduces `-aux-range "(START STOP)"`, which resolves the selected symbols to runtime addresses and configures perf so that:

- Intel PT starts paused
- hitting START resumes AUX tracing
- hitting STOP pauses AUX tracing

The control events are emitted in the same perf event group as the Intel PT source.

## Motivation

magic-trace currently relies on userspace snapshot/control paths for trigger-based workflows. perf's AUX pause/resume support allows tracing to be gated directly by perf events, avoiding unnecessary trace collection outside the selected region.

## Implementation

- add typed AUX actions (`start-paused`, `resume`, `pause`)
- add breakpoint-backed AUX control events
- resolve AUX range symbols using the existing ELF/runtime address path
- support both `run` and `attach`
- reject AUX ranges when stacktrace sampling is selected
- add serialization and generated-perf-argument expect tests

## Validation

- full tests cannot run on macOS because the locked dependency graph includes Linux-only Jane Street dependencies

Opening as draft to validate the full Linux build/test suite before marking ready.

### DIFF
--- a/src/aux_trace_range.ml
+++ b/src/aux_trace_range.ml
@@ -1,0 +1,23 @@
+open! Core
+
+type t =
+  { start_symbol : Symbol_selection.t
+  ; stop_symbol : Symbol_selection.t
+  }
+
+let param =
+  let open Command.Param in
+  flag
+    "-aux-range"
+    (optional
+       (Arg_type.create (fun s ->
+          let start_symbol, stop_symbol =
+            Sexp.of_string s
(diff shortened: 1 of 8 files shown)


namtran1812:
Draft for validation: DCO is passing, but the Linux build workflow is currently blocked before any jobs are created, so I haven't received a compile/test result yet. I'll keep this in draft until CI and Intel PT validation are complete.

PR #387, opened 2026-08-15
Handle trigger breakpoints across process threads

### LINKED ISSUE #274: stop indicator does not work in multi-threaded application
We have a simple reproducer:

```cpp
///usr/bin/env -S clang++ -O0 "$0" -o /tmp/demo && exec /tmp/demo "$@"
#include <iostream>
#include <fstream>
#include <thread>

using namespace std;


extern "C" {
void __attribute__((noinline)) __attribute__((used)) magic_trace_stop_indicator() { }
}

void task() {
    ifstream r;
    ofstream w;
    r.open("/dev/zero", ios::binary);
    w.open("/dev/null", ios::binary);
    char buf[4096] = {0};
    for (int i = 0; i >= 0; i++)
    {
        r.read(buf, sizeof(buf));
        w.write(buf, sizeof(buf));
        if (i % 100 == 0) {
            magic_trace_stop_indicator();
        }
    }
}

int main(int argc, char *argv[])
{
    if (argc > 1) {
        std::cout << "running in thread" << std::endl;
        std::thread t1(task);
        t1.join();
    } else {
        std::cout << "running in main" << std::endl;
        task();
    }
}
```

`magic_trace_stop_indicator` can be triggered reliably in main thread, but not in child thread.

namtran1812:
## Summary

Fix trigger breakpoints so they can fire from worker threads when magic-trace is attached to a multithreaded process.

Fixes #274.

## Problem

The existing trigger implementation creates a hardware breakpoint only for the head PID/TID. Since `perf_event_open` breakpoints are task-specific, a trigger such as `magic_trace_stop_indicator()` can work from the main thread but be missed when called from a worker thread.

## Implementation

- Enumerate process TIDs from `/proc/<pid>/task`
- Introduce a `Trigger_breakpoint` abstraction for breakpoint lifecycle and Async FD monitoring
- Maintain one trigger breakpoint per discovered TID
- Rescan for threads created after attachment
- Handle races where threads exit during breakpoint setup
- Preserve single-snapshot behavior across thread breakpoints
- Add a multithreaded regression target based on #274

## Validation

- multithreaded C++ regression target builds successfully

End-to-end perf breakpoint behavior requires Linux validation, so this is being opened as a draft.

### DIFF
--- a/src/process_info.ml
+++ b/src/process_info.ml
@@ -26,4 +26,27 @@ let read_all_proc_info () =
     | _ -> ())
 ;;
 
+let thread_ids_of_dir_entries entries =
+  Array.filter_map entries ~f:(fun entry ->
+    try Some (Pid.of_string entry) with
+    | _ -> None)
+  |> Array.to_list
+  |> List.sort ~compare:Pid.compare
+;;
+
+let thread_ids pid =
+  Or_error.try_with (fun () ->
+    Sys_unix.readdir [%string "/proc/%{pid#Pid}/task"] |> thread_ids_of_dir_entries)
+;;
+
+let thread_exists ~pid ~tid =
+  match Core_unix.access [%string "/proc/%{pid#Pid}/task/%{tid#Pid}"] [ `Exists ] with
+  | Ok () -> true
+  | Error _ -> false
+;;
+
+module For_testing = struct
+  let thread_ids_of_dir_entries = thread_ids_of_dir_entries
+end
+
 let cmdline_of_pid pid = Hashtbl.find state pid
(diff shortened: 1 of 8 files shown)


namtran1812:
Draft for validation: DCO is passing, but the Linux build workflow is currently blocked before any jobs are created. I’ll keep this in draft until CI and end-to-end multithreaded perf validation are complete.

namtran1812:
Hi! I noticed the Actions workflow for this PR expired while waiting for approval from a maintainer, so the build/tests never ran. Would you mind approving or re-running CI when you have a chance? Happy to address anything that comes up. Thanks!

PR #388, opened 2026-08-15
Keep per-thread trace timestamps monotonic

### LINKED ISSUE #319: "duration_complete event must have start tick (166428285) greater than end tick (166428276)"
When running `./magic-trace run -working-directory workdir -full-execution -multi-thread ./dotnet/dotnet -- benchmark/Abc.TradingPerfTests.dll`  

magic-trace end up throwing when writing the trace with the error:
`"duration_complete event must have start tick (166428285) greater than end tick (166428276)"`  

Thanks for helping us make magic-trace better!

Please consider answering any subset of the following questions, based on what you think
is most relevant to your report:

> - How can we, magic-trace developers, reproduce the issue you're seeing?

If needed I can provide a zip that contains everything to reproduce the issue.

> - What version of perf userspace tools are you running? `perf --version`

`perf version 6.8.12`

- What CPU do you have? `cat /proc/cpuinfo`  

`Intel(R) Xeon(R) Gold 6146 CPU`

namtran1812:
## Summary

Prevent backwards per-thread event timestamps from producing invalid trace durations during decoding.

Fixes #319.

## Problem

The streaming trace writer assumes that event time for a given thread is monotonic, but that invariant was not enforced.

When an event arrives with a timestamp older than one already processed for the same thread:

- `last_event_time` moves backwards
- pending-event flushing can operate over a negative time interval
- open callstack frames can later be closed at a timestamp earlier than their start

This can ultimately produce errors such as:

`duration_complete event must have start tick (...) greater than end tick (...)`

## Implementation

Normalize incoming per-thread event timestamps at the `event_time` boundary:

- preserve stream processing order
- keep the writer's logical per-thread time monotonic
- prevent pending-event flushing from operating over negative intervals
- keep callstack reconstruction and thread-end timestamps on the same ordered timeline

The fix intentionally lives at timestamp normalization rather than clamping individual duration writes.

## Tests

Added a deterministic regression case with:

- call at 200 ns
- nested call at 300 ns
- return event at 150 ns

The trace writer must serialize this stream successfully instead of producing an invalid backwards duration.

## Validation

- full inline tests cannot currently run in the local macOS switch because Jane Street test-only dependencies are unavailable there

Opening as draft pending upstream Linux CI validation.

### DIFF
--- a/src/trace_writer.ml
+++ b/src/trace_writer.ml
@@ -506,6 +506,11 @@ let event_time t (event : Event.t) (thread_info : _ Thread_info.t) =
       thread_info.last_event_time
     | Some time ->
       let time = map_time t time in
+      (* perf can occasionally emit events for a thread with timestamps older than an
+         event we have already processed. Keep the writer's per-thread logical time
+         monotonic so pending-event flushing and callstack reconstruction never operate
+         over a negative time interval. *)
+      let time = Mapped_time.max time thread_info.last_event_time in
       thread_info.last_event_time <- time;
       time
   in
--- a/test/test.ml
+++ b/test/test.ml
@@ -196,6 +196,24 @@ let dump_using_file ?range_symbols events =
   return ()
 ;;
 
+let write_using_file events =
+  let%bind events = get_events_pipe ~events () in
+  let close_result = return (Ok ()) in
+  let buf = Iobuf.create ~len:500_000 in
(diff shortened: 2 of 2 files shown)


namtran1812:
Hi! I noticed the Actions workflow for this PR expired while waiting for approval from a maintainer, so the build/tests never ran. Would you mind approving or re-running CI when you have a chance? Happy to address anything that comes up. Thanks!

